PR #8, opened 2020-10-07
Add cpp-linenoise as an external dependency for the fuzzer REPL

octurion:


WORKSPACE
```diff
@@ -18,6 +19,14 @@ http_archive(
     urls = ["https://github.com/google/googletest/archive/release-1.10.0.zip"],
 )
 
+new_git_repository(
```
werat, line 22: Let's use `http_archive` instead, this is the best practice -- https://docs.bazel.build/versions/master/external.html#repository-rules

build_defs/cpp_linenoise.BUILD
```diff
@@ -0,0 +1,20 @@
...
+        "-fno-exceptions",
+        "-fno-rtti",
+    ],
+})
+
+cc_library(
+    name = "cpp_linenoise",
+    hdrs = ["linenoise.hpp"],
+    includes = ["."],
+)
```
werat, line 20: Let's add `include_prefix="cpp_linenoise"` to make it more "contained" 
https://docs.bazel.build/versions/master/be/c-cpp.html#cc_library.include_prefix

Then in the code it will be `#include "cpp_linenoise/linenoise.hpp"`.

build_defs/cpp_linenoise.BUILD
```diff
@@ -17,4 +17,5 @@ cc_library(
     name = "cpp_linenoise",
     hdrs = ["linenoise.hpp"],
     includes = ["."],
+    include_prefix = "linenoise",
```
werat, line 20: Nit, to make it clear about the library name
```suggestion
    include_prefix = "cpp_linenoise",
```

werat (approved):

PR #12, opened 2020-10-09
Refactor project structure

werat:
Move the library code src -> lldb-eval.
Refactor includes and BUILD rules.


lldb-eval/BUILD
```diff
@@ -0,0 +1,92 @@
...
+    deps = [
+        ":lldb-eval",
+        ":runner",
+        "@com_google_googletest//:gtest",
+        "@com_google_googletest//:gtest_main",
+        "@llvm_project//:lldb-api",
+    ],
+)
+
+cc_test(
```
octurion, line 66: Maybe it would be a good idea to keep the testcases in a separate directory/separate directories?

tools/fuzzer/BUILD
```diff
@@ -0,0 +1,58 @@
...
+    deps = [
+        ":fuzzer_lib",
+        "//lldb-eval",
+        "@bazel_tools//tools/cpp/runfiles",
+        "@io_github_yhirose_cpplinenoise//:cpp_linenoise",
+        "@llvm_project//:lldb-api",
+    ],
+)
+
+cc_test(
```
octurion, line 47: Ditto about testcases and separate directories.

octurion (commented):
Only question/suggestion is whether or not to split unit testing into a separate directory/separate directories. Other than that, LGTM.

lldb-eval/BUILD
werat: With Bazel and gtest it's common to have `lib.cc` and `lib_test.cc` next to each other. IMO it's a good practice (at least for now), it keeps the test local and easy to discover.

PR #13, opened 2020-10-09
Add types support to fuzzer

octurion:


tools/fuzzer/ast.h
```diff
@@ -25,6 +25,120 @@
 
 namespace fuzzer {
 
+enum class ScalarType : unsigned char;
+class TaggedType;
+class PointerType;
+class ReferenceType;
+
+using QualifiableType = std::variant<ScalarType, TaggedType, PointerType>;
```
werat, line 33: Should `ReferenceType` also be qualifiable?

tools/fuzzer/ast.h
```diff
@@ -268,6 +383,23 @@ class TernaryExpr {
...
+
+  const Type& type() const;
+  const Expr& expr() const;
+  int precedence() const { return PRECEDENCE; }
+
+  friend std::ostream& operator<<(std::ostream& os, const CastExpr& expr);
+
+ private:
+  Type type_;
+  std::unique_ptr<Expr> expr_;
```
werat, line 400: Why `expr_` is a pointer, but `type_` is not?

tools/fuzzer/expr_gen.h
```diff
@@ -42,6 +42,50 @@ enum class ExprKind : unsigned char {
...
+
+  using expr_iterator = decltype(expr_weights_)::iterator;
+  using expr_const_iterator = decltype(expr_weights_)::const_iterator;
+
+  using type_iterator = decltype(type_weights_)::iterator;
+  using type_const_iterator = decltype(type_weights_)::const_iterator;
+
+ public:
+  expr_iterator expr_begin() { return expr_weights_.begin(); }
+  expr_iterator expr_end() { return expr_weights_.end(); }
```
werat, line 67: I feel like operating on iterators is kinda clunky. Can you just expose the array itself instead? Then you can do range for when you don't need the index.

tools/fuzzer/expr_gen.h
```diff
@@ -61,6 +105,9 @@ struct GenConfig {
 
   float parenthesize_prob = 0.5f;
```
werat, line 106: Side note, we should probably decrease this


4ir6zr

tools/fuzzer/fuzzer_test.cc
```diff
@@ -351,4 +434,83 @@ std::vector<TestParam> gen_params() {
...
 
-INSTANTIATE_TEST_SUITE_P(Fuzzer, OperatorPrecedence, ValuesIn(gen_params()));
+INSTANTIATE_TEST_SUITE_P(Fuzzer, OperatorPrecedence,
+                         ValuesIn(gen_precedence_params()));
+
+struct TypePrintTestParam {
+  std::string str;
+  std::shared_ptr<Type> type;
+
+  explicit TypePrintTestParam(std::string str, Type type)
```
werat, line 444: nit: doesn't have to be `explicit`

tools/fuzzer/ast.h
octurion: No, `int& *` is not a valid C++ type, hence the differentiation between `Type` (allows refs) and `QualifiableType` (doesn't allow refs).

tools/fuzzer/ast.h
octurion: Because `Expr` already contains `CastExpr` as a possible variant type, hence a definition of `Expr expr_` would mean that `CastExpr` is of infinite size.

werat (approved):

PR #14, opened 2020-10-12
Make fuzzer use the runner, refactor runner

octurion:


lldb-eval/runner.cc
```diff
@@ -69,65 +69,78 @@ int FindBreakpointLine(const Runfiles& runfiles,
...
-          break;
-        }
-        default:
-          break;
-      }
-
-    } else {
+  auto listener = debugger.GetListener();
+
+  for (;;) {
```
werat, line 102: Prefer `while(true)` for infinite loops, `for(;;)` is kinda cryptic.
```suggestion
  while (true) {
```

lldb-eval/runner.cc
```diff
@@ -69,65 +69,78 @@ int FindBreakpointLine(const Runfiles& runfiles,
...
+    auto stopReason = thread.GetStopReason();
+
+    if (stopReason != lldb::eStopReasonBreakpoint) {
+      continue;
+    }
+
+    auto bpId =
+        static_cast<lldb::break_id_t>(thread.GetStopReasonDataAtIndex(0));
+    if (bpId != bp.GetID()) {
+      std::cerr << "Stopped at unknown breakpoint: " << bpId << std::endl;
```
werat, line 138: I think here we should also exit with an error. If the process is stopped at a different breakpoint, no one is going to resume it.

werat (approved):

PR #16, opened 2020-10-12
Implement bitwise not (~) and unary negation (!)

werat:


lldb-eval/eval_test.cc
```diff
@@ -221,6 +221,22 @@ TEST_F(InterpreterTest, TestArithmetic) {
   TestExpr("-20LL / 1ULL", "18446744073709551596");
 }
 
+TEST_F(InterpreterTest, TestBitwiseOperators) {
+  TestExpr("~(-1)", "0");
+  TestExpr("~~0", "0");
+  TestExpr("~0", "-1");
+  TestExpr("~1", "-2");
```
jaro-sevcik, line 228: Could you also test this or 64-bit numbers, bools and something that is not a scalar?

octurion (approved):

PR #17, opened 2020-10-14
Add configuration struct for fuzzer

octurion:


tools/fuzzer/expr_gen.cc
```diff
@@ -157,22 +144,52 @@ Expr ExprGenerator::maybe_parenthesized(Expr expr) {
...
+  for (size_t i = 0; i < mask.size(); i++) {
+    if (running_ones == choice && mask[i]) {
+      return (UnOp)i;
+    }
+    if (mask[i]) {
+      running_ones++;
+    }
+  }
+
+  // Fallback, should never be reached
```
werat, line 190: Can you explain here why it's not reachable? E.g. "choice is always less than mask.count(), so we're guaranteed to have running_ones == choice at some point".

tools/fuzzer/expr_gen.cc
```diff
@@ -157,22 +144,52 @@ Expr ExprGenerator::maybe_parenthesized(Expr expr) {
...
+
+  size_t running_ones = 0;
+  for (size_t i = 0; i < mask.size(); i++) {
+    if (running_ones == choice && mask[i]) {
+      return (UnOp)i;
+    }
+    if (mask[i]) {
+      running_ones++;
+    }
+  }
```
werat, line 188: Alternative way to compute this. I'm not sure which one is easier to understand (mine seems a bit more intuitive to me, but dunno), so up to you to decide.

```suggestion
  std::uniform_int_distribution<size_t> distr(1, mask.count());
  size_t choice = distr(rng_);

  size_t running_ones = 0;
  for (size_t i = 0; i < mask.size(); i++) {
    if (mask[i]) {
      running_ones++;
    }
    if (running_ones == choice) {
      return (UnOp)i;
    }
  }
```

tools/fuzzer/expr_gen.h
```diff
@@ -40,26 +41,65 @@ enum class ExprKind : unsigned char {
...
+
+  uint64_t int_const_min = 0;
+  uint64_t int_const_max = 1000;
+
+  double double_constant_min = 0;
+  double double_constant_max = 10;
+
+  float parenthesize_prob = 0.5f;
+
+  BinOpMask bin_op_mask = ~0ull ^ ((1ull << (unsigned long long)BinOp::Shl) |
```
werat, line 63: Can you do something like this?

```
BinOpMask().flip()
           .flip(BinOp::Shl)
           ...
           .flip(BinOp::Shr);
```

Idk if it will look nice after `clang-format` tho.

werat (approved):

PR #21, opened 2020-10-16
Add boolean constants to fuzzer

octurion:


tools/fuzzer/ast.h
```diff
@@ -268,6 +269,29 @@ class TernaryExpr {
...
 };
 
+class BooleanConstant {
+ public:
+  enum class Value : bool { False, True };
+
+  static constexpr int PRECEDENCE = 0;
+
+  static constexpr Value True = Value::True;
+  static constexpr Value False = Value::False;
```
werat, line 279: Is this a shortcut to write `BooleanConstant::True` instead of `BooleanConstant::Value::True`? If so, I think it breaks the logic a bit. One would expect `BooleanConstant::True` to return a `BooleanConstant` that represents `True`, not the underlying `Value`.

tools/fuzzer/ast.h
```diff
@@ -268,6 +269,29 @@ class TernaryExpr {
...
+
+  Value value() const { return value_; }
+  int precedence() const { return PRECEDENCE; }
+
+ private:
+  Value value_;
+};
+
+std::ostream& operator<<(std::ostream& os, BooleanConstant::Value value);
+
```
werat, line 294: Do you really need this indirection? Can you just print `BooleanConstant` directly (same as you're doing to all other nodes)?

werat (approved):

PR #25, opened 2020-10-22
Make `unreachable()` not depend on `llvm_unreachable()`

octurion:


werat (approved):

lldb-eval/eval.h
```diff
@@ -20,7 +20,6 @@
 #include "clang/Basic/TokenKinds.h"
 #include "expression_context.h"
 #include "lldb-eval/ast.h"
-#include "lldb-eval/defines.h"
```
werat, line 23: Add this include to all files that use `lldb_eval_unreachable`, please. Ideally every symbol usage should have an appropriate include for it (it may not be the case currently, but we try to adhere to this rule).

lldb-eval/parser_test.cc
```diff
@@ -20,8 +20,6 @@
 #include "lldb-eval/expression_context.h"
 #include "lldb/API/SBExecutionContext.h"
 
-// DISALLOW_COPY_AND_ASSIGN is also defined in
-// lldb/lldb-defines.h
```
werat, line 24: Why removing this comment? We're still including LLDB headers here, so this comment is relevant, no?

octurion: Oops, I misread it as `lldb-eval/defines.h` 😬

werat (approved):

PR #28, opened 2020-10-28
Werat/value refactor

werat:
This refactoring removes `Scalar` and `Pointer` types, moving all relevant operations to `Value`. Value now is the one true way of storing and moving around "values".

Value now holds precise type information about the underlying value, which is represented as `lldb::SBValue`. Before Scalar was losing precise information (i.e. `int` vs `long` or `long` vs `long long`, depending on the platform).


lldb-eval/value.cc
```diff
@@ -14,279 +14,392 @@
...
+
+    } else {
+      lldb_eval_unreachable("Invalid arithmetic type");
+    }
+
+  } else if (other.IsInteger()) {
+    // The value being promoted can be only an integer, no floats allowed here.
+
+    // TODO(werat): It's possible that `other` should be promoted too, if the
+    // it's type can't represent all values of `promo_candidate`. For example,
```
jaro-sevcik: the it's => its

jaro-sevcik (approved):
Looks good once comments are addressed.

Thanks!

PR #29, opened 2020-10-29
Use WIN32_LEAN_AND_MEAN to speedup the build

werat:


octurion (commented):
Would we benefit at all from defining `WIN32_LEAN_AND_MEAN`, though? Asking because I'm not seeing any inclusion of `windows.h` from a quick grep (in which case WIN32_LEAN_AND_MEAN would make sense).

FYI, I also had a look at `VC_EXTRALEAN` and it does not seem to be of use to us.

werat:
It is included indirectly from LLVM. But it seems it doesn't really bring any speed up (according to CI timings).

PR #31, opened 2020-11-02
Resolve identifiers during parsing

werat:
Resolve the identifiers during parsing and stored the resolved values in
the AST. This makes it easier to report correct errors and saves time on
evaluating invalid expressions. Ideally, most of the check should be
done during parsing (type resolution, type checking, etc).


lldb-eval/parser.cc
```diff
@@ -522,6 +532,7 @@ ExprResult Parser::ParsePostfixExpression() {
       case clang::tok::plusplus:
       case clang::tok::minusminus: {
         BailOut(
+            ErrorCode::kNotImplemented,
             "Don't support postfix inc/dec yet: " + TokenDescription(token_),
             token_.getLocation());
```
octurion, line 537:
```suggestion
            llvm::formatv("We don't support postfix inc/dec yet: {0}",
                TokenDescription(token_)),
            token_.getLocation());
```

lldb-eval/parser.cc
```diff
@@ -561,10 +570,31 @@ ExprResult Parser::ParsePrimaryExpression() {
...
     return ParseBooleanLiteral();
   } else if (token_.isOneOf(clang::tok::coloncolon, clang::tok::identifier)) {
-    return ParseIdExpression();
+    // Save the source location for the diagnostics message.
+    clang::SourceLocation loc = token_.getLocation();
+    auto identifier = ParseIdExpression();
+    auto value = expr_ctx_->LookupIdentifier(identifier.c_str());
+    if (!value) {
+      BailOut(ErrorCode::kUndeclaredIdentifier,
+              "use of undeclared identifier '" + identifier + "'", loc);
```
octurion, line 579:
```suggestion
              llvm::formatv("use of undeclared identifier '{0}', identifier), loc);
```

lldb-eval/parser_test.cc
```diff
@@ -28,30 +28,30 @@
 
 namespace {
```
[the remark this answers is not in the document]
octurion: So I had a look around and looks like the gdb people have fixed this: https://sourceware.org/bugzilla/show_bug.cgi?id=16874

However, I also spotted e.g. https://youtrack.jetbrains.com/issue/CPP-18948 (which is related to CLion), so even with the gdb fix in mind, I would still go with `static` instead of an anonymous namespace.

octurion (approved):

PR #37, opened 2020-11-16
Implement pointer comparison operations

werat:
* Support pointer literal (nullptr)
* Add pointer <-> integer comparisons


lldb-eval/eval_test.cc
```diff
@@ -492,6 +492,40 @@ TEST_F(EvalTest, TestPointerArithmetic) {
...
+  EXPECT_THAT(Eval("nullptr == 0"), IsEqual("true"));
+  EXPECT_THAT(Eval("0 != nullptr"), IsEqual("false"));
+
+  EXPECT_THAT(
+      Eval("(void*)0 > nullptr"),
+      IsError(
+          "invalid operands to binary expression ('void *' and 'nullptr_t')"));
+
+  // These are not allowed by C++, but we support it as an extension.
+  EXPECT_THAT(Eval("(void*)1 == 1"), IsEqual("true"));
```
jaro-sevcik, line 515: Do we have tests for comparing non-zero void* to another non-zero void*?

jaro-sevcik (approved):

lldb-eval/eval_test.cc
werat: There are a few, but I've added more.

lldb-eval/eval_test.cc
```diff
@@ -492,6 +484,55 @@ TEST_F(EvalTest, TestPointerArithmetic) {
...
+  EXPECT_THAT(Eval("(void*)1 == 0"), IsEqual("false"));
+  EXPECT_THAT(Eval("(void*)1 > 0"), IsEqual("true"));
+  EXPECT_THAT(Eval("(void*)1 < 0"), IsEqual("false"));
+  EXPECT_THAT(Eval("1 > (void*)0"), IsEqual("true"));
+  EXPECT_THAT(Eval("2 < (void*)3"), IsEqual("true"));
+
+  // Integer is converted to uintptr_t, so negative numbers because large
+  // positive numbers.
+  EXPECT_THAT(Eval("(void*)0xffffffffffffffff == -1"), IsEqual("true"));
```
octurion, line 531:
```suggestion
  EXPECT_THAT(Eval("(void*)0xffff'ffff'ffff'ffff == -1"), IsEqual("true"));
```

I have a feeling this might not be 100% portable. It certainly works on 32-bit and 64-bit machines, but what if 128-bit machines become a thing? (Then again, this might take a long time 😛).

PR #40, opened 2020-11-30
Support bitfields

werat:
ReadRawData doesn't handle bitfield constraints, so use
SBValue::GetValueAsUnsigned for reading integral values.


testdata/test_binary.cc
```diff
@@ -413,6 +413,23 @@ void TestValueScope() {
   // BREAK(TestValueScope)
 }
 
+void TestBitField() {
+  struct BitFieldStruct {
+    uint16_t a : 10;
+    uint32_t b : 4;
+    bool c : 1;
```
octurion, line 420: In C/C++, it is possible to enforce bit field alignment as follows:
```
unsigned char foo : 8;
bool bar : 1;
unsigned char : 0;
uint16_t quux: 16; // Not placed in the same "storage unit" as foo and bar
```

Maybe we should add a testcase for this?

werat: Added more tests

PR #42, opened 2020-12-04
Add tests for static const members

werat:
Some of the tests are disabled for now, since the requred patches are
not yet submitted to LLDB.


testdata/test_binary.cc
```diff
@@ -282,16 +282,48 @@ int i = 2;
...
+  static const unsigned int static_const;
 };
 
-const int Foo::y = 42;
+const unsigned int Vars::static_const = 3;
 
-static void TestQualifiedId() {
-  // BREAK(TestQualifiedId)
+}  // namespace inner
+
```
octurion, line 303: I would also suggest adding anonymous namespaces as a testcase.

werat: Anonymous namespaces are actually not supported right now. Variables (and types) declared in anonymous namespaces are unique for compilation units, so it's not clear how to do a proper lookup. Potentially we could try to find the variable from the current CU, but that might be expensive.

octurion (commented):
I guess we should address #44 before merging.

octurion (approved):

PR #43, opened 2020-12-07
Fuzzer improvements, implement pointer dereference, address-of, array indexing

octurion:
* Implement pointer dereference, address-of, and array indexing.
* Allow RNG seed to be passed as a command line argument.
* Fuzzer output layout improvements, print errors from lldb and lldb-eval as well.
* Fix incorrect precedence for cast expression.
* Implement symbol table and pick an appropriate variable at random.
* Fix constraints bug where for a constraint that would allow type `T`, creation of a constraint `T*`  would also allow `void*`.
* Improve the fake RNG used for unit testing so that expressions picking a random type implicitly (e.g. ternary expression) can be tested.
* Specify a maximum depth for the fuzzer.

Previous commit (since its commit message is blank):
* Make all AST nodes copy-constructible and hashable.
* Add dereference expression.
* Add EnumBitset type.
* Allow fuzzer to backtrack.


tools/fuzzer/expr_gen.h
```diff
@@ -122,8 +125,19 @@ struct GenConfig {
       {1.0f, 0.1f},  // TypeKind::PointerType
       {1.0f, 0.1f},  // TypeKind::VoidPointerType
   }};
+};
+
+/*
+ * The purpose for generating the type in question. The purpose of this enum
+ * is purely to assist in unit testing.
+ */
+enum class GenTypePurpose {
```
werat, line 134: It's quite sad that something required just for testing changes the "real" code so much.

It seems, at the moment it's only needed for deducing the result type of the ternary operator. How about we just add the `result_type_` field to the `TernaryExpr`? Then in tests you can just propagate it like this and avoid all this implicit/explicit type picking:

```
// clang-format off
Expr expected = UnaryExpr(
    UnOp::BitNot,
    ParenthesizedExpr(TernaryExpr(
        ScalarType::SignedInt,
        IntegerConstant(1),
        IntegerConstant(2),
        IntegerConstant(3))));
    // clang-format on
```

This would also make sense for other constructs, where you need to deduce result type.

tools/fuzzer/ast.cc
```diff
@@ -168,8 +168,17 @@ BinaryExpr::BinaryExpr(Expr lhs, BinOp op, Expr rhs)
     : lhs_(std::make_shared<Expr>(std::move(lhs))),
       rhs_(std::make_shared<Expr>(std::move(rhs))),
       op_(op) {}
+BinaryExpr::BinaryExpr(Expr lhs, BinOp op, Expr rhs, Type type_for_unit_testing)
```
werat, line 171: Let's call it just `result_type` or `expr_type`. It's not just for unit testing, it makes total sense to have it in the AST.

werat (approved):

PR #45, opened 2020-12-07
Resolve the type in C-style cast during parsing

werat:
... and store the resolved SBType in the AST node. This allows to
produce better diagnostics and to re-use the type during the evaluation.


lldb-eval/parser.cc
```diff
@@ -1025,19 +1073,45 @@ void Parser::ParsePtrOperator(TypeDeclaration* type_decl) {
...
+      }
+      // Get pointer type for the base type: e.g. int* -> int**.
+      type = type.GetPointerType();
+
+    } else if (tk == clang::tok::amp) {
+      // References to references are forbidden.
+      if (type.IsReferenceType()) {
+        BailOut(ErrorCode::kInvalidOperandType,
+                "type name declared as a reference to a reference",
+                token_.getLocation());
```
jaro-sevcik, line 1106: It looks like the error is a bit different from the above, is it intentional?

If not, then you can perhaps hoist the `if (type.IsReferenceType()) {` the check to the beginning of the loop.

jaro-sevcik (approved):

lldb-eval/parser.cc
werat: The error messages are the same as in `clang`.  Also these cases are different: one is pointer-to-reference, second is reference-to-reference.

PR #46, opened 2020-12-07
Better output format, allow only mismatches/errors to be output

octurion:


tools/fuzzer/main.cc
```diff
@@ -41,8 +41,99 @@
...
+    value_mismatch = lldb_value.GetValue() != lldb_eval_value.GetValue();
+  }
+
+  bool type_mismatch;
+  if (lldb_value.GetTypeName() != nullptr &&
+      lldb_eval_value.GetTypeName() != nullptr) {
+    type_mismatch =
+        strcmp(lldb_value.GetTypeName(), lldb_eval_value.GetTypeName()) != 0;
+  } else {
+    // Pointer comparison: Mismatch if one type is null and the other is not
```
werat, line 77: Not 100%, but I've never seen `GetTypeName` to return `nullptr`

werat (approved):

tools/fuzzer/main.cc
octurion: Yep, had a look at the lldb source code and it's using [https://lldb.llvm.org/cpp_reference/ConstString_8h_source.html](ConstString) under the hood, which doesn't look like it could be null (since the default ctor initializes it to an empty string).

octurion: Nope, I spotted a case where lldb-eval returned a value with an empty type, so the extra test will have to stay for now.

PR #52, opened 2021-01-05
Support context variables

tonkosi:
lldb-eval can't use context variables (aka. convenience variables) out of the box. In order to support context variables in lldb-eval, API methods are extended to explicitly take a identifier->value map as an argument.


lldb-eval/eval_test.cc
```diff
@@ -94,6 +95,39 @@ class EvaluatorHelper {
...
+  EvalResult EvalWithContext(
+      const std::string& expr,
+      const std::unordered_map<std::string, lldb::SBValue>& vars) {
+    EvalResult ret;
+
+    std::vector<lldb_eval::ContextVariable> ctx_vec;
+    ctx_vec.reserve(vars.size());
+    for (const auto& it : vars) {
+      ctx_vec.push_back({it.first.c_str(), it.second});
+    }
```
werat, line 107:
```suggestion
    for (const auto& [name, value] : vars) {
      ctx_vec.emplace_back(name.c_str(), value);
    }
```

lldb-eval/eval_test.cc
```diff
@@ -1061,3 +1116,71 @@ TEST_F(EvalTest, TestBitField) {
...
+  EXPECT_THAT(Scope("s").EvalWithContext("*$ptr", vars_), IsEqual("'h'"));
+  EXPECT_THAT(Scope("s").EvalWithContext("$ptr[1]", vars_), IsEqual("'e'"));
+
+  EXPECT_THAT(EvalWithContext("$ctx + *$ptr", vars_), IsEqual("117"));
+  EXPECT_THAT(Scope("s").EvalWithContext("$ctx + *$ptr", vars_),
+              IsEqual("117"));
+
+  // Context variables that were created in this test are visible from LLDB.
+  // Disable comparisons with LLDB to test subsets of created context variables.
+  this->compare_with_lldb_ = false;
```
werat, line 1157: Can you put this and everything below in a separate `TEST_F` function? As you've noticed `compare_with_lldb_` changes the behavior for the current test, so ideally we'd like to minimize the scope.

lldb-eval/eval_test.cc
tonkosi: Done, also implemented constructor for `ContextVariable`.

lldb-eval/eval_test.cc
tonkosi: Done.

lldb-eval/api.h
```diff
@@ -45,6 +45,9 @@ namespace lldb_eval {
 struct ContextVariable {
   const char* name;
   lldb::SBValue value;
+
+  ContextVariable(const char* name, lldb::SBValue value)
+      : name(name), value(value) {}
```
werat, line 50: Oh, I didn't realize `emplace_back` requires a user-defined ctor. Exported inline functions are evil (ctor in this case) -- https://devblogs.microsoft.com/oldnewthing/20140109-00/?p=2123, let's try to avoid them. I'd just revert is to `push_back`, doesn't make much difference. Sorry for confusion :)

tonkosi: Reverted. I learned something new, too :)

werat (approved):

PR #54, opened 2021-01-18
Do semantic analysis during parsing

werat:
Move most of the semantic analysis from the evaluation stage to the parsing stage. This allows to resolve the types of all sub-expressions before evaluation and perform necessary type promotions and conversions. Namely, the result type of the ternary operator depends on both of the operands, so it's not possible to do it correctly during evaluation. Doing semantic analysis during parsing also improves diagnostic error messages, addind arrowed context.

This change introduces one new major (and potentially dangerous) invariant -- the AST produced by the parser is assumed to be valid, unless parser reported an error. Evaluation of this AST must be always successful and can produce only runtime errors (e.g. null pointer dereference).

This change also fixes a few minor bugs (and adds test cases), which were uncovered during the semantic analysis implementation.


lldb-eval/parser.cc
```diff
@@ -158,13 +168,288 @@ lldb::BasicType PickIntegerType(const clang::NumericLiteralParser& literal,
...
+
+  assert(CompareTypes(lhs->result_type_deref(), rhs->result_type_deref()) &&
+         "operands result types must be the same");
+
+  return lhs->result_type_deref().GetCanonicalType();
+}
+
+static lldb::SBTypeMember GetFieldWithNameIndexPath(
+    lldb::SBType type, const std::string& name, std::vector<uint32_t>* idx) {
+  // Direct base classes are location before fields, so field members needs to
```
jaro-sevcik, line 411: location -> located

lldb-eval/parser.cc
```diff
@@ -158,13 +168,288 @@ lldb::BasicType PickIntegerType(const clang::NumericLiteralParser& literal,
...
+  // be offsetted by the number of base classes.
```
jaro-sevcik, line 412: offsetted -> offset

lldb-eval/parser.cc
```diff
@@ -1323,4 +1616,698 @@ ExprResult Parser::ParseIntegerLiteral(clang::NumericLiteralParser& literal,
...
+      if (type.GetByteSize() < sizeof(void*)) {
+        BailOut(ErrorCode::kInvalidOperandType,
+                llvm::formatv(
+                    "cast from pointer to smaller type '{0}' loses information",
+                    type.GetName()),
+                location);
+        return std::make_unique<ErrorNode>();
+      }
+    }
+    // Otherwise accept only arithmetic types and enums.
```
jaro-sevcik, line 1647: Nit: Perhaps put the comment inside the then block. Otherwise, it looks like the previous 'if' was already finished. (Here and below.)

jaro-sevcik (approved):
Looks great, I only have a few questions and small suggestions.

lldb-eval/value.cc
```diff
@@ -86,27 +84,117 @@ lldb::SBType GetEnumerationIntegerType_V(T type, lldb::SBTarget target) {
...
+}
+
+bool Type::IsPointerToVoid() {
+  return IsPointerType() &&
+         GetPointeeType().GetBasicType() == lldb::eBasicTypeVoid;
+}
+
+bool Type::IsNullPtrType() { return GetBasicType() == lldb::eBasicTypeNullPtr; }
+
+bool Type::IsSigned() { return GetTypeFlags() & lldb::eTypeIsSigned; }
```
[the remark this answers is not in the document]
werat: Hmm, not sure, everything seems to work fine without calling `GetCanonicalType`. I guess it's left after previous refactorings.

PR #57, opened 2021-01-26
Implement evaluation flow analysis

werat:
Keep track of the current evaluation context and propagate the relevant state to sub-expressions. This allows to eliminate "pair" operations like "&*".

Also remove all occurences of `lldb_eval_unreachable` from the Interpreter and remove the `Error` argument, since we're not producing meaningful diagnostics anyway.


lldb-eval/eval.h
```diff
@@ -79,8 +93,11 @@ class Interpreter : Visitor {
   // types, create objects, perform casts, etc.
   lldb::SBTarget target_;
 
+  // Flow analysis contains the evaluation information about the current code
+  // branch. It allows to eliminate some "pair" operations (e.g. `&*`).
+  std::vector<FlowAnalysis*> flow_analysis_stack_;
```
jaro-sevcik, line 98: Instead of using an explicit stack, you could just chain the on-stack flow analysis objects in a linked list (each flow analysis object would have a parent pointer).

lldb-eval/eval_test.cc
```diff
@@ -653,6 +653,23 @@ TEST_F(EvalTest, TestPointerDereference) {
...
+  EXPECT_THAT(Eval("**pp_int0"), IsEqual("0"));
+  EXPECT_THAT(Eval("**pp_int0 + 1"), IsEqual("1"));
+  EXPECT_THAT(Eval("&**pp_int0"), IsOk());
+  EXPECT_THAT(Eval("&**pp_int0 + 1"), IsOk());
+
+  EXPECT_THAT(Eval("&(true ? *p_null : *p_null)"),
+              IsEqual("0x0000000000000000"));
+  EXPECT_THAT(Eval("&(false ? *p_null : *p_null)"),
+              IsEqual("0x0000000000000000"));
+  EXPECT_THAT(Eval("&*(true ? p_null : nullptr)"),
```
jaro-sevcik, line 671: So what happens when you do &&**nullptr?

lldb-eval/eval.h
```diff
@@ -79,8 +93,11 @@ class Interpreter : Visitor {
   // types, create objects, perform casts, etc.
   lldb::SBTarget target_;
 
+  // Flow analysis contains the evaluation information about the current code
+  // branch. It allows to eliminate some "pair" operations (e.g. `&*`).
```
jaro-sevcik, line 97: Please add a more detailed description of this means - what are the elements of the vector and what null means.

jaro-sevcik (commented):
The idea looks great. There are some things that bother me a bit:

- What is the meaning of null

lldb-eval/eval.h
werat: Done

lldb-eval/eval.h
werat: The problem is that flow analysis objects are not always created (which may be a problem on it's own), so this chaining will be a bit tricky.

werat:
> The idea looks great. There are some things that bother me a bit:
> 
> * The mutable nature of FlowAnalysis. Please describe somewhere how the algorithm works, it is very confusing.
> * The semantics of flow_analysis_stack_ is a bit unclear.

I've written a more detailed explanation about the flow analysis. I think it still might be a bit confusing, but at least it's used only in a few places for now.

> * As far I can see, the is-address-of-pending only makes in some of the nodes. It would be great to have some assertions that is false in nodes where it does not make sense.

Well, `is-address-of-pending` can still be true in any of the nodes, it's just it doesn't apply there. E.g. if you have `&a + 1`, the `IdentifierNode` representing `a` will have `is-address-of-pending = true`. And things like `&1` are forbidden by the parser.

lldb-eval/eval_test.cc
jaro-sevcik: Is not that bad? I thought &* pairs should cancel out.

I was wondering if it would not be more understandable if the evaluation methods returned pair of dereference-count and value, then you would get the actual value by applying dereference dereference-count-times on the value.

Address-of would be then calculated as simply decreasing the dereference count. If you needed the real value (e.g., to do arithmetic), you would just apply the dereferences.

lldb-eval/eval.cc
```diff
@@ -172,43 +173,45 @@ static Value EvaluateArithmeticOp(lldb::SBTarget target,
     return EvaluateArithmeticOpFloat(target, kind, lhs, rhs, rtype);
   }
 
-  lldb_eval_unreachable("Result of arithmetic conversion is not a scalar");
+  return Value();
```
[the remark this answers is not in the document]
jaro-sevcik: Better, thanks.

lldb-eval/eval.h
```diff
@@ -93,9 +93,16 @@ class Interpreter : Visitor {
   // types, create objects, perform casts, etc.
   lldb::SBTarget target_;
 
-  // Flow analysis contains the evaluation information about the current code
-  // branch. It allows to eliminate some "pair" operations (e.g. `&*`).
-  std::vector<FlowAnalysis*> flow_analysis_stack_;
+  // Flow analysis chain represents the expression evaluation flow for the
+  // current code branch. Each node in the chain corresponds to an AST node,
+  // describing the semantics of the evaluation for it. For example, flow
```
jaro-sevcik, line 98: "For example" -> "Currently"

Or does it do anything else?

jaro-sevcik (approved):

jaro-sevcik (approved):

lldb-eval/eval_test.cc
werat: `& &**p` is not valid in C++, "&*" pair should not be eliminated in this case.

At first I tried to implement by doing the dereference lazily, i.e. have `Value` to know about pending derefs and doing it only when the value is actually requested. But that get tricky in other cases. For example, `&(*foo_ptr)->bar` should resolve to the `foo_ptr + offset_of(bar)` :/

PR #58, opened 2021-01-29
Force static values

werat:
By default `target.prefer-dynamic-value=no-run-target`, which means LLDB will try to resolve the actual object type for dynamic types. This can break the evaluation, since the parser relies only on static types.


jaro-sevcik (approved):
Looks great, thanks!

PR #59, opened 2021-02-02
Support ">>" for nested templates

werat:
... and C++-style casts, when those are implemented.


tonkosi (approved):
Looks good to me.

I'm just curious about templated type name being parsed to string. For example, how are currently handled types `T_1<ns::T_1>` and `T_1<::ns::T_1>`, are these treated as different types?
Are there any obstacles when comparing a parsed templated type to LLDB type?

werat:
> I'm just curious about templated type name being parsed to string. For example, how are currently handled types `T_1<ns::T_1>` and `T_1<::ns::T_1>`, are these treated as different types?

These two _can_ be treated different types, depending on the context. Basically, we parse every type argument separately, try to resolve it and theт construct the final type in a string form. `::ns::T_1` is a fully qualified name and thus unambiguous, but `ns::T_1` can be resolved to different types depending on the current context.

> Are there any obstacles when comparing a parsed templated type to LLDB type?
Yep, our type should be _exactly_ the same as it's stored in the index.

PR #61, opened 2021-02-04
Allow casting arrays to pointers

werat:


lldb-eval/eval_test.cc
```diff
@@ -1071,6 +1071,13 @@ TEST_F(EvalTest, TestCStyleCastNullptrType) {
   EXPECT_THAT(Eval("(char*)nullptr"), IsEqual("0x0000000000000000"));
 }
 
+TEST_F(EvalTest, TestCStyleCastArray) {
+  EXPECT_THAT(Eval("(int*)arr"), IsOk());
+  EXPECT_THAT(Eval("(char*)arr"), IsOk());
+  EXPECT_THAT(Eval("((char*)arr)[0]"), IsEqual("'\\x01'"));
+  EXPECT_THAT(Eval("((char*)arr)[1]"), IsEqual("'\\0'"));
+}
```
tonkosi, line 1079: Maybe add test with casting multi-dimensional arrays to pointer types?
At first I thought casting from 2d arrays may not work because assigning arrays to pointers works only for 1d arrays, but it turns out there's no problem with casting 😃

tonkosi (approved):
It looks good, I just have a small suggestion regarding tests.

lldb-eval/eval_test.cc
werat: Thanks for the suggestion, can't have too many tests here :)

PR #62, opened 2021-02-05
Support C++ style casts

werat:
Introduce basic support for C++ style casts: `static_cast`, `dynamic_cast` and `reinterpret_cast`.

`dynamic_cast` is not supported by LLDB's expression evaluator, so we disable it too for now.

`static_cast` and `reinterpret_cast` are currently redirected to C-style cast.


jaro-sevcik (approved):
Thanks!

PR #63, opened 2021-02-05
Improve memory access constraints in fuzzer

tonkosi:
This change tightens memory access control when creating expressions by keeping a track of number of times an expression can be dereferenced ("freedom index"). Before this change, a single bool was used to determine whether an expression must be a valid pointer, but that lead to issues with expressions such as `**ptr`, where `ptr` is a valid pointer that can only be dereferenced once.

In addition, a small change to comparing LLDB and lldb-eval types is introduced that ignores type qualifiers to reduce unwanted noise produced when running the fuzzer.


tools/fuzzer/constraints_test.cc
```diff
@@ -176,3 +176,33 @@ TEST(Constraints, Unsatisfiability) {
   EXPECT_THAT(default_ctor.satisfiable(), IsFalse());
   EXPECT_THAT(default_specific_types.satisfiable(), IsFalse());
 }
+
+TEST(Constraints, MemoryConstraints) {
+  auto expect_fields = [](const MemoryConstraints& constraints,
```
werat, line 181: Maybe you can just define `operator==` for `MemoryConstraints` (here in the test file) and then just use EXPECT_EQ:

```
bool operator==(const MemoryConstraints& lhs, const MemoryConstraints& rhs){ 
    return lhs.must_be_valid() == rhs.must_be_valid() &&
        lhs.required_freedom_index() == rhs.required_freedom_index();
}
...
EXPECT_EQ(invalid.from_address_of(), {false, 0});
```

werat (commented):
Thanks, looks great! One small suggestion regarding the tests.

tools/fuzzer/constraints_test.cc
tonkosi: Nice suggestion, that should improve precision of test reports.
However, the change wasn't as smooth as expected :/ Using `{false, 0}` in `EXPECT_EQ` resulted in compile error because of macro limitations (apparently, wrapping `{}` expression in parens seems to be a common workaround, but didn't work in this case). Even `EXPECT_EQ(lhs, MemoryConstraint(false, 0))` didn't compile.
I decided to use the form `EXPECT_TRUE(lhs == MemoryConstraints(false, 0))`. For constructors I check each field explicitly (to avoid possible modifications by constructor).

werat (approved):

tools/fuzzer/constraints_test.cc
tonkosi: `EXPECT_EQ(lhs, MemoryConstraints(false, 0))` works after all. Equality comparator just had to be part of the `fuzzer` namespace.

PR #64, opened 2021-02-09
Perform array-to-pointer conversion

werat:
Arrays are implicitly converted to pointers during usual arithmetic conversions.


lldb-eval/parser.cc
```diff
@@ -252,6 +252,13 @@ static ExprResult UsualUnaryConversions(lldb::SBTarget target,
 
   // TODO(werat): Promote bitfields: e.g. uint32_t:10 -> int (not unsigned int).
 
+  if (result_type.IsArrayType()) {
```
tonkosi, line 255: Nit: can you update the comment few lines above saying that this only includes the integral promotion?

tonkosi (approved):
LGTM with one nit.

PR #65, opened 2021-02-09
Add source locations for each ptr declarator

werat:
This makes error messages to point exactly to the offending declarator.


tonkosi (approved):
Looks good, thanks.

PR #68, opened 2021-02-12
Fix casting enums to floating point types

tonkosi:
Before this change, reinterpreted cast was used for casting enums to basic types, leading to errors when casting to floating point types (e.g. `(double)Enum::kOne` was close to `4E-324`). This change fixes this by using a static cast instead.


werat (approved):
Awesome, thanks for fixing this!

PR #69, opened 2021-02-15
Implement builtin debugger intrinsic `__log2`

werat:
Implement builtin function in the parser and the evaluator. Currently only one function is supported -- `__log2`.

* Save source location in the AST node to allow for better error diagnostics
* Implement basic implicit type conversions checking -- this is required for handling function arguments


docs/expr-lang.ebnf
```diff
@@ -140,3 +141,12 @@ numeric_literal = ? clang::tok::numeric_constant ? ;
 boolean_literal = "true" | "false" ;
 
 pointer_literal = "nullptr" ;
+
+builtin_func = builtin_func_name "(" {builtin_func_argument_list} ")" ;
+
+builtin_func_name = "__log2" ;
+
+builtin_func_argument_list = builtin_func_argument
+                           | builtin_func_argument_list "," builtin_func_argument_list
```
tonkosi, line 150: Is this supposed to be `builtin_func_argument "," builtin_func_argument_list`?

lldb-eval/parser.cc
```diff
@@ -1780,7 +1816,102 @@ ExprResult Parser::ParseIntegerLiteral(clang::NumericLiteralParser& literal,
...
+  ConsumeToken();
+
+  Expect(clang::tok::l_paren);
+  ConsumeToken();
+
+  std::vector<ExprResult> arguments;
+
+  do {
+    // Eat the comma if this is not the first iteration.
+    if (arguments.size() > 0) {
```
jaro-sevcik, line 1850: Nit: I guess you could consume the token unconditionally and remove the ConsumeToken for lparen.

docs/expr-lang.ebnf
jaro-sevcik, line 150: One of those builtin_func_argument_lists could be builtin_func_argument, no?

Also, is there a reason to forbid empty argument list?

lldb-eval/eval_test.cc
```diff
@@ -1618,3 +1618,41 @@ TEST_F(EvalTest, TestSizeOf) {
...
+  EXPECT_THAT(Eval("__log2(cxx_enum)"), IsEqual("7"));
+  EXPECT_THAT(Eval("__log2(CEnum::kFoo)"), IsEqual("7"));
+  EXPECT_THAT(Eval("__log2(CxxEnum::kFoo)"), IsEqual("7"));
+
+  EXPECT_THAT(Eval("__log2(foo)"),
+              IsError("no known conversion from 'Foo' to 'unsigned int'\n"
+                      "__log2(foo)\n"
+                      "       ^"));
+
+  EXPECT_THAT(Eval("1 + __log2(1, 2)"),
```
jaro-sevcik, line 1653: Also add a test for no arguments, please.

lldb-eval/parser.cc
```diff
@@ -938,6 +966,7 @@ ExprResult Parser::ParseUnaryExpression() {
   }
 
   if (token_.is(clang::tok::kw_sizeof)) {
+    clang::Token token = token_;
```
tonkosi, line 969: Nit: The source location can be calculated directly here.

lldb-eval/eval_test.cc
jaro-sevcik: Perhaps also for something >2^32.

docs/expr-lang.ebnf
jaro-sevcik: (Looking at the parser, it looks like non-empty arguments was an intentional decision.)

lldb-eval/eval_test.cc
jaro-sevcik: And maybe the usual double suspects (+/- infinity, nan).

lldb-eval/parser.cc
```diff
@@ -1780,7 +1816,102 @@ ExprResult Parser::ParseIntegerLiteral(clang::NumericLiteralParser& literal,
...
+
+    arguments.push_back(std::move(argument));
+
+  } while (token_.is(clang::tok::comma));
+
+  Expect(clang::tok::r_paren);
+  ConsumeToken();
+
+  // Check we have the correct number of arguments.
+  if (arguments.size() != func_def->arguments_.size()) {
```
tonkosi, line 1869: We could also bail out a bit earlier in the case there are more arguments than expected with additional check in the do-while loop. Is that going to make a difference?

docs/expr-lang.ebnf
werat: Technically the other way around (`builtin_func_argument_list "," builtin_func_argument`), but yes, thanks for catching this!

lldb-eval/parser.cc
werat: Done

lldb-eval/parser.cc
werat: This is no longer the case when handling empty argument list

lldb-eval/parser.cc
werat: We need to parse all the arguments so that we can have an error message like "requires N arguments, got K arguments".

tonkosi: Makes sense, thanks for clarification.

tonkosi (approved):

PR #73, opened 2021-03-01
Support static_cast and reinterpret_cast in fuzzer

tonkosi:
This change supports C++ style casts in fuzzer.
It also disables generation and use of qualified types by default. The reason is sensitivity of casting away type qualifiers in C++ casts, which currently isn't supported in the fuzzer. When qualified types are enabled, C++ casts are disabled.


werat (approved):
Looks great!

PR #74, opened 2021-03-03
Add base for generation of function calls

tonkosi:
This change implements `FunctionCallExpr` node. It is tested with `__log2`, currently the only function supported by lldb-eval. Unfortunately, `__log2` isn't supported by LLDB, so this expression kind won't be used until there's a function supported by both LLDB and lldb-eval.

### DIFF
--- a/tools/fuzzer/fuzzer_test.cc
+++ b/tools/fuzzer/fuzzer_test.cc
@@ -156,6 +156,13 @@ class FakeGeneratorRng : public GeneratorRng {
     return fuzzer::Field(std::move(type), std::move(field_name));
   }
 
+  const fuzzer::Function& pick_function(
+      const std::vector<std::reference_wrapper<const fuzzer::Function>>&)
+      override {
+    assert(false && "Not implemented yet!");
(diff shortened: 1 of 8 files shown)


tools/fuzzer/fuzzer_test.cc
```diff
@@ -156,6 +156,13 @@ class FakeGeneratorRng : public GeneratorRng {
     return fuzzer::Field(std::move(type), std::move(field_name));
   }
 
+  const fuzzer::Function& pick_function(
+      const std::vector<std::reference_wrapper<const fuzzer::Function>>&)
+      override {
+    assert(false && "Not implemented yet!");
+    return fuzzer::Function("dummy", {});
```
werat, line 163: Returning reference to a temporary value. Either create this dummy function at a global/class scope, or just return by value. I think for our purposes copying `Function` object is not that bad.

tonkosi: Replaced returning by reference with returning by value.

werat (commented):
Looks great!

werat (approved):

PR #75, opened 2021-03-05
Implement increment and decrement

werat:
Implement prefix and postfix increment and decrement operations. These operations mutate have side-effects (i.e. they mutate the memory), therefore they're allowed only for context variables. Context variables are used specifically to maintain state between the expressions and they are (generally) not in the target process memory.


lldb-eval/eval_test.cc
```diff
@@ -1696,3 +1696,61 @@ TEST_F(EvalTest, TestBuiltinFunction_Log2) {
       Eval("::ns::dummy(1)"),
       IsError("function '::ns::dummy' is not a supported builtin intrinsic"));
 }
+
+TEST_F(EvalTest, TestPrefixIncDec) {
```
tonkosi, line 1700: I noticed that LLDB allows `++(++$x)`, but `lldb-eval` results with `expression is not assignable`.

tonkosi (changes requested):
Overall it looks good, but it seems that prefix increment/decrement is an L-value in C++, i.e. `++x` is assignable in LLDB, but not currently in lldb-eval.

lldb-eval/eval_test.cc
werat: It is allowed in C++ (and we can actually implement it too), but MSVC debugger doesn't allow such side-effects:
```
Natvis: C:\Users\werat\src\cpp\Dummy\NatvisFile.natvis(19,17): Error: Side effects are not supported in this context.
    Error while evaluating '++(++index)' in the context of type 'Dummy.exe!Foo'.
```

So for now we can get away without it too. I've added a comment about this.

tonkosi: Thanks for the comment and clarification.

PR #77, opened 2021-03-09
Fix lookup for global pointer and reference identifiers

tonkosi:
LLDB API's `SBValue::GetName()` sometimes include a variable's type, which makes identifier lookup a bit trickier. `lldb-eval` already tries to match name by ignoring everything in front of the last occurrence of space (e.g. for names in the form of `"const int var"`), but that isn't sufficient for pointer and reference types where `*` and `&` symbols are adjacent to the name (e.g. `"int *ptr"`, `"int &ref"`). This change solves this by adding checks to ignore everything in front of the last occurrence of `*` or `&`.


werat (approved):
It's terrible we have to resort to such hacks, but we do what we have to :)

PR #79, opened 2021-03-11
[NFC] Use custom enum for binary operation kind

werat:
Using `clang::tok::TokenKind` was a quick-hack in the early days and now it's time to be more explicit.


tonkosi (approved):
Looks great!

PR #80, opened 2021-03-12
Support array indirection

tonkosi:


werat (approved):
Great, thanks for fixing this!

PR #81, opened 2021-03-15
Use arrays in bool context

tonkosi:


lldb-eval/value.cc
```diff
@@ -188,6 +189,8 @@ bool Value::IsFloat() { return type_.IsFloat(); }
 
 bool Value::IsPointer() { return type_.IsPointerType(); }
 
+bool Value::IsArray() { return type_.IsArrayType(); }
```
werat, line 192: This method is used only inside `Value`, so let's just inline it. Most of `Value::Is*` are here for convenience, because doing `value.type().Is*` is annoying.

werat (commented):
Thanks, looking good! Please, rebase on master, I've fixed the Windows CI check.

werat (approved):

PR #85, opened 2021-03-23
In value scope `this` should refer to the object itself

werat:


lldb-eval/eval_test.cc
```diff
@@ -1347,6 +1347,21 @@ TEST_F(EvalTest, TestValueScope) {
...
+  // Test for the "artificial" value, i.e. created by the expression.
+  lldb::SBError error;
+  lldb::SBValue scope_var =
+      lldb_eval::EvaluateExpression(frame_, "(test_scope::Value&)bytes", error);
+  EXPECT_TRUE(scope_var.IsValid());
+  EXPECT_TRUE(error.Success());
+
+  EvaluatorHelper scope(scope_var, true);
+  EXPECT_THAT(scope.Eval("this->y_"), IsEqual("2.5"));
+  EXPECT_THAT(scope.Eval("(*this).y_"), IsEqual("2.5"));
```
tonkosi, line 1363: Maybe add a test to make sure that `this` is undeclared when evaluated in the frame context?

tonkosi (approved):
LGTM with a suggestion regarding tests.

lldb-eval/eval_test.cc
werat: That's a good idea, done!

PR #87, opened 2021-03-24
Fix casting and comparison with nullptr

tonkosi:
This change includes:

* Implement support for zero literal and use it to compare nullptrs (`nullptr == 0`)
* Implement casting to `nullptr_t`


lldb-eval/parser.cc
```diff
@@ -1822,11 +1825,19 @@ ExprResult Parser::ParseIntegerLiteral(clang::NumericLiteralParser& literal,
 
   auto [type, is_unsigned] = PickIntegerType(literal, raw_value);
 
+  bool is_zero_literal =
```
werat, line 1828: How about something like this?

```c++
bool is_zero_literal = raw_value.isNullValue() && value.IsInteger();
```

Also, what are other options for `type`? Can it be anything else apart from the "integer"?

lldb-eval/parser.cc
```diff
@@ -1988,6 +1999,17 @@ ExprResult Parser::BuildCStyleCast(Type type, ExprResult rhs,
     }
     kind = CStyleCastKind::kPointer;
 
+  } else if (type.IsNullPtrType()) {
+    // Cast to nullptr type.
+    if (!rhs_type.IsNullPtrType() && !rhs->is_literal_zero()) {
+      BailOut(ErrorCode::kInvalidOperandType,
+              llvm::formatv("cannot cast from type '{0}' to type '{1}'",
```
werat, line 2006: Is this the message `clang` would give in this context? I've tried `(std::nullptr_t)1` and got:
```
<source>:8:14: error: C-style cast from 'int' to 'std::nullptr_t' (aka 'nullptr_t') is not allowed
    auto x = (std::nullptr_t)1;
             ^~~~~~~~~~~~~~~~~
```

lldb-eval/parser.cc
tonkosi: Thanks for the suggestion, that looks much cleaner!
I can't think of anything else apart "integer" (`false` and real numbers can't be used in this context), except for `#define`s to literal zero (e.g. `NULL`).

Just "integer" also isn't the best option, e.g. char literal `'\0'` cannot be used in this context, but char literals aren't supported currently, so I added a `TODO`.

lldb-eval/parser.cc
tonkosi: That's right, I updated the message.

Also, note that `static_cast<std::nullptr_t>(1)` results with the same "C-style" message, that should be fixed when we improve C++ casts handling.

lldb-eval/parser.cc
werat: I mean this is `Parser::ParseIntegerLiteral`, which parses _only_ integer literals. Boolean, floats, chars and so on are handled in other places. So it seems to me here `value.IsInteger()` is guaranteed to be true (e.g. `PickIntegerType` returns only integers types by design).

tonkosi: Totally forgot about that. I reduced it to just `raw_value.isNullValue()`.

werat (approved):
Looks great, thanks!

PR #90, opened 2021-03-31
Use dereferenced type in usual unary conversions

tonkosi:
This change fixes the issue when using references to arrays and enums in arithmetic operations (before they are implicitly converted to pointer and integer types respectively).

Examples of evaluations that failed before this change:
```
int (&arr_ref)[3] = arr;
> arr_ref + 1
Invalid operands to binary expression ('int [3]' and 'int')

CEnum& c_enum_ref = c_enum;
> c_enum_ref - 1
Assertion failed...
```


werat (approved):
Great, thanks for fixing this!

PR #91, opened 2021-04-01
Detect member function calls

werat:
Member function calls are function calls too and should be classified by the parser as such.


tonkosi (approved):
Looks good!

PR #92, opened 2021-04-06
Improve type checking in ternary expression

tonkosi:
This improves use of arrays and literal `0` in ternary expression:

* `condition ? array : pointer` works fine
* `condition ? pointer : 1` fails as expected


lldb-eval/parser.cc
```diff
@@ -2709,25 +2709,69 @@ ExprResult Parser::BuildTernaryOp(ExprResult cond, ExprResult lhs,
...
 
+  // At this point the resulting type should be a pointer, and at least one
+  // operand should either be an array or a pointer.
+
+  // Array operand types should be converted to pointer types.
+  auto lhs_type_orig = lhs_type;
+  auto rhs_type_orig = rhs_type;
+  if (lhs_type.IsArrayType()) {
+    lhs_type = lhs_type.GetArrayElementType().GetPointerType();
+  }
```
werat, line 2737: I think we can do `array->pointer` conversion right here, i.e.

```
if (lhs_type.IsArrayType()) {
  lhs  = std::make_unique<CStyleCastNode>(...);
}
```

After this we just need to check if `lhs` and `rhs` have the same pointer type (or apply nullptr/zero -> pointer conversion).

werat (changes requested):

lldb-eval/parser.cc
tonkosi: The reasons why I didn't convert it here are because I'd still need to do conversion for nullptr/zero later (so we have this only in one place) and to avoid allocating new memory in the case we don't have to (in the "incompatible operand types" case).

Anyway, I guess that extra allocation is negligible for performance and if you think it is more readable that way, I don't mind :)

lldb-eval/parser.cc
```diff
@@ -2709,25 +2709,69 @@ ExprResult Parser::BuildTernaryOp(ExprResult cond, ExprResult lhs,
...
+  // Deduce the resulting pointer type.
+  lldb::SBType pointer_type;
+  if (lhs_type.IsPointerType()) {
+    pointer_type = lhs_type;
+  } else if (rhs_type.IsPointerType()) {
+    pointer_type = rhs_type;
+  }
+
+  // Is `type` convertible to the `pointer_type`?
+  auto is_valid_type = [&pointer_type](Type type) {
```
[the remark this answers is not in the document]
tonkosi: Done.

tonkosi (commented):
Also added few tests with enums.

lldb-eval/parser.cc
```diff
@@ -2709,25 +2709,71 @@ ExprResult Parser::BuildTernaryOp(ExprResult cond, ExprResult lhs,
...
+  if (lhs_type.IsArrayType()) {
+    lhs_type = lhs_type.GetArrayElementType().GetPointerType();
+    lhs = std::make_unique<CStyleCastNode>(
+        lhs->location(), lhs_type, std::move(lhs), CStyleCastKind::kPointer);
+  }
+  if (rhs_type.IsArrayType()) {
+    rhs_type = rhs_type.GetArrayElementType().GetPointerType();
+    rhs = std::make_unique<CStyleCastNode>(
+        rhs->location(), rhs_type, std::move(rhs), CStyleCastKind::kPointer);
+  }
```
werat, line 2744: Actually it seems that according to the C++ standard array-to-pointer conversion happens before everything else. This doesn't change much, but if we do that we can simplify this method even more.

If we apply `array-to-pointer` in the beginning before `if (CompareTypes(...))`, we can handle the case when both operands are the same array there. As a bonus, this will also fix a corner case we currently have -- the result type of `cond ? arr1 : arr2` should be _pointer_, but at the moment we'll have array.

And then later we can do just two more checks:
```
if (lhs_type.IsPointerType() && (rhs_type.IsNullPtrType() || rhs->is_literal_zero()) {
    rhs = std::make_unique<CStyleCastNode>(...)
    return std::make_unique<TernaryOpNode>(...);
}
if ((lhs_type.IsNullPtrType() || lhs->is_literal_zero() && rhs_type.IsPointerType() ) {
    lhs = std::make_unique<CStyleCastNode>(...)
    return std::make_unique<TernaryOpNode>(...);
}
```

I would put them before the case for `nullptr <-> 0`, as it likely a lot more common.

lldb-eval/parser.cc
```diff
@@ -2709,25 +2709,60 @@ ExprResult Parser::BuildTernaryOp(ExprResult cond, ExprResult lhs,
...
+  // If one operand is nullptr and the other one is literal zero, convert
+  // the literal zero to a nullptr type.
+  if (lhs_type.IsNullPtrType() && rhs->is_literal_zero()) {
+    rhs = std::make_unique<CStyleCastNode>(
+        rhs->location(), lhs_type, std::move(rhs), CStyleCastKind::kNullptr);
+
+    return std::make_unique<TernaryOpNode>(location, lhs_type, std::move(cond),
+                                           std::move(lhs), std::move(rhs));
+  }
+  if (rhs_type.IsNullPtrType() && lhs->is_literal_zero()) {
```
werat, line 2758:
```suggestion
  if (lhs->is_literal_zero() && rhs_type.IsNullPtrType()) {
```

werat (commented):
Looks great, thanks!

werat (approved):

PR #93, opened 2021-04-07
[fuzzer] Disable auto fix-its

tonkosi:


werat (commented):
Can you also disable them in `tools/exec` please?

werat (approved):
Thanks!

PR #111, opened 2021-04-23
Document list of recommended patches and known bugs in LLDB

werat:


tonkosi (approved):
Looks good, thanks for documenting this.

PR #114, opened 2021-05-03
Allow enabling/disabling side effects

werat:
Allow setting expression evaluation options via `lldb_eval::Options`. Introduce an option `allow_side_effects` that controls whether side effects are allowed during the expression evaluation.


lldb-eval/eval_test.cc
```diff
@@ -2224,3 +2238,22 @@ TEST_F(EvalTest, TestCompositeAssignmentBitwise) {
       EvalWithContext("$p >> $p", vars_),
```
tonkosi, line 2238: Not related to this change, but was this supposed to be `>>=`? There are also 5 other occurrences of `>>` or `<<` above in this test.

tonkosi (approved):
Looks great! The only comment I have is regarding some older tests.

lldb-eval/eval_test.cc
werat: Oh, whoops, that's a typo. Good catch!

PR #116, opened 2021-05-05
Support member access on struct arrays

tonkosi:


lldb-eval/parser.cc
```diff
@@ -2893,15 +2893,19 @@ ExprResult Parser::BuildMemberOf(ExprResult lhs, std::string member_id,
...
                             "you mean to use '.'?",
                             lhs_type.GetName()),
               location);
       return std::make_unique<ErrorNode>();
     }
-    lhs_type = lhs_type.GetPointeeType();
+    if (lhs_type.IsPointerType()) {
+      lhs_type = lhs_type.GetPointeeType();
+    } else {
+      lhs_type = lhs_type.GetArrayElementType();
```
werat, line 2907: If `lhs` is an array, we should insert an `array-to-pointer` conversion here. This way the interpreter doesn't need to do anything.

tonkosi: Done

werat (approved):

PR #117, opened 2021-05-10
[fuzzer] Simplify type constraints

tonkosi:
There used to be two classes for type constraints: `SpecificTypes` and `TypeConstraints` (a wrapper for `SpecificTypes`). These classes are merged into a single `TypeConstraints` class in this change.


werat (commented):
It looks a lot simpler now, thanks!

werat (approved):

PR #118, opened 2021-05-10
Move array-to-pointer conversion to InsertArrayToPointerConversion

werat:
Insert array-to-pointer decay conversion for the dereference and the array subscript operations to simplify the logic in the `Interpreter`.


lldb-eval/eval.cc
```diff
@@ -355,21 +355,13 @@ void Interpreter::Visit(const ArraySubscriptNode* node) {
     return;
   }
 
-  assert((base.type().IsPointerType() || base.type().IsArrayType()) &&
-         "array subscript: base must be pointer or array");
+  assert((base.type().IsPointerType()) &&
```
tonkosi, line 358: This brackets are no longer needed.

tonkosi (approved):
I have one tiny comment, otherwise it looks good :)

PR #119, opened 2021-05-11
[fuzzer] Improve type satisfiability checks

tonkosi:
Before this change it could happen that the expression generator takes unsatisfiable type constraints, so the generator couldn’t produce any kind of expression. But instead of stopping early, the generator would continue to try to generate a valid expression, exhausting all possible options. As depth was increasing, there were even more options which turned out to be invalid. As a result of combinatorial explosion, generation with depth limit of just `15` would often take almost a second to complete.

This change:
* fixes a few bugs that causes unsatisfiable constraints,
* adds an assertion to make sure type constraints are always satisfied (otherwise, there’s probably a bug somewhere else),
* allows leaf expression kinds when depth limit is reached.

To demonstrate the performance impact of this change, I ran a small benchmark in which I measured the time and node count in the generation tree (`GenNode`s and random values) while calling `ExprGenerator::generate()` 10000 times (depth limit = 15):
* before:
  * Time (total): ~460 seconds
  * Node count (avg): ~16000
  * Node count (max): ~300000
* after:
  * Time (total): ~13 seconds
  * Node count (avg): ~420
  * Node count (max): ~90000


tools/fuzzer/constraints.cc
```diff
@@ -228,7 +228,9 @@ TypeConstraints TypeConstraints::allowed_to_point_to() const {
   }
 
   if (std::holds_alternative<AnyType>(ptr_types_)) {
-    return TypeConstraints::all();
+    TypeConstraints retval = TypeConstraints::all();
+    retval.scalar_types_[ScalarType::Void] = false;
+    return retval;
```
werat, line 233: Should we also remove `void` from `specific_types_ptr`?

tools/fuzzer/expr_gen.cc
```diff
@@ -628,6 +626,14 @@ std::optional<Expr> ExprGenerator::gen_address_of_expr_impl(
 
   TypeConstraints new_type_constraints =
       constraints.type_constraints().allowed_to_point_to();
+
+  if (!new_type_constraints.satisfiable()) {
+    // It is possible that type constraints become unsatisfiable after the
+    // `allowed_to_point_to()` is performed (e.g. `void*` can't become `void`
+    // since `void` expressions are invalid).
```
werat, line 633: Does this already happen or is it introduced in this pull request? IIUC right now `allowed_to_point_to` doesn't do the pruning of `void` from `void*` (it only forbids `void` if pointer constraint is `AnyType`).

werat (approved):

PR #121, opened 2021-05-12
Add a workflow to run benchmarks for pull requests

werat:
The workflows triggers for pull requests, run the benchmarks for the baseline (base revision) and for the contender (head revision from the pull requests), compares the results and posts them as a comment to the pull requests.


.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,76 @@
...
+
+          # Use python3 to run the compare script, because numpy already dropped
+          # support for python2. The compare script fails with the current
+          # version of numpy because of the non-ASCII symbol in the source code:
+          # https://github.com/numpy/numpy/blob/6d7b8aaed5ae9f0435764675ebac8c9ada06738f/numpy/__init__.py#L292
+          sed -i 's/PY2/PY3/g' tools/BUILD.bazel
+
+          bazel run //tools:compare -- --no-color benchmarks ${{ env.BASELINE_JSON }} ${{ env.CONTENDER_JSON }} | tee cmp_results
+
+          echo 'Benchmark comparison for [`${{ env.BASE_SHA }}`](${{ github.event.repository.html_url }}/commit/${{ github.event.pull_request.base.sha }}) (base) vs [`${{ env.HEAD_SHA }}`](${{ github.event.repository.html_url }}/commit/${{ github.event.pull_request.head.sha }}) (PR)' >> pr_comment
```
tonkosi, line 60: It would be cool if this was formatted as a markdown table, but let's not complicate it now :)

tonkosi (approved):
Looks good, thanks for setting this up!

.github/workflows/benchmarks.yml
werat: Yeah, that would be nice :) The compare script can actually output the results as JSON (via `--dump_to_json` flag), so it's possible to format the results in a nicer way. It would be also cool to colorize the output (e.g. green for improvements and red for degradations) and add some warning for significant changes.

PR #123, opened 2021-05-17
Support looking up enum values in symbol table test

werat:
Enum values can be looked up as global variables if the corresponding LLDB patch is available (https://reviews.llvm.org/D94077). Update `PopulateSymbolTableTest.Variables` to support this behaviour.


tonkosi (approved):
Looks good, thanks for fixing it.

PR #125, opened 2021-05-18
Always use type size from the target

werat:
Don't use `static_cast<long>(value)` to do type casts, because the size of `long` on the host might be different from the target. Always use the type size from the target process.


lldb-eval/value.cc
```diff
@@ -285,72 +286,107 @@ void Value::Update(Value v) {
...
-    break;                                               \
+static llvm::APFloat CreateAPFloatFromAPSInt(const llvm::APSInt& value,
+                                             lldb::BasicType basic_type) {
+  switch (basic_type) {
+    case lldb::eBasicTypeFloat:
+      return llvm::APFloat(value.isSigned()
+                               ? llvm::APIntOps::RoundSignedAPIntToFloat(value)
+                               : llvm::APIntOps::RoundAPIntToFloat(value));
+    case lldb::eBasicTypeDouble:
+      // No way to get more precision at the moment.
```
tonkosi, line 297: Did `clang-format` add this extra indentation? This way it may seem that the comment corresponds to the `lldb::eBasicTypeDouble` instead of `lldb::eBasicTypeLongDouble`. Would it make more sense to move it under `lldb::eBasicTypeLongDouble` or inline it with it?

lldb-eval/value.cc
```diff
@@ -285,72 +286,107 @@ void Value::Update(Value v) {
...
-
-  switch (type.GetCanonicalType().GetBasicType()) {
-#define CASE(basic_type, builtin_type)                   \
-  case basic_type: {                                     \
-    auto v = static_cast<builtin_type>(val.GetUInt64()); \
-    ret = CreateValueFromBytes(target, &v, type);        \
-    break;                                               \
-  }
+Value CastEnumToBasicType(lldb::SBTarget target, Value val, Type type) {
+  assert(type.IsScalar() && "target type must be an scalar");
```
tonkosi, line 364:
```suggestion
  assert(type.IsScalar() && "target type must be a scalar");
```

tonkosi (approved):
Looks great, thanks for fixing it!

lldb-eval/value.cc
werat: The extra indent is indeed from `clang-format`. I agree it looks a bit confusing, but other options seem even more confusing to me :)

PR #126, opened 2021-05-19
[fuzzer] Add libfuzzer target

tonkosi:


tools/fuzzer/BUILD
```diff
@@ -106,3 +107,27 @@ cc_test(
...
+        # not necessary on Windows, but "tags" attribute is not configurable
+        # with select -- https://github.com/bazelbuild/bazel/issues/2971.
+        "no-sandbox",
+    ],
+    srcs = ["lldb_eval_libfuzzer_test.cc"],
+    deps = [
+        "//lldb-eval",
+        "//lldb-eval:runner",
+        "@bazel_tools//tools/cpp/runfiles",
+        ":fuzzer_lib",
```
werat, line 129: Also needs `"@llvm_project//:lldb-api"`

tonkosi: Added. I think this was also missing in `symbol_table_test` and `mutate_test`, so added there too.

tools/fuzzer/lldb_eval_libfuzzer_test.cc
```diff
@@ -0,0 +1,145 @@
...
+  std::ostringstream os;
+  os << maybe_expr.value();
+
+  lldb::SBError error;
+  lldb_eval::EvaluateExpression(g_frame, os.str().c_str(), error);
+  return 0;
+}
+
+extern "C" int LLVMFuzzerInitialize(int* /*argc*/, char*** /*argv*/) {
+  g_runfiles = Runfiles::CreateForTest();
```
werat, line 129: I'm not sure about the regular `bazel test`, but in oss-fuzz this target is executed as a "normal" binary. There no additional env vars set by Bazel, so `CreateForTest` won't work.

I had to use `Runfiles::Create((*argv)[0])`.

tools/fuzzer/lldb_eval_libfuzzer_test.cc
```diff
@@ -0,0 +1,163 @@
...
+  os << maybe_expr.value();
+
+  lldb::SBError error;
+  lldb_eval::EvaluateExpression(g_frame, os.str().c_str(), error);
+  return 0;
+}
+
+extern "C" int LLVMFuzzerInitialize(int* argc, char*** argv) {
+  assert(*argc >= 1);
+  g_runfiles = Runfiles::Create((*argv)[0]);
```
werat, line 147: Runfiles is used only during the initialization, so it doesn't have to be a global variable. This also avoid a memory leak, since currently `g_runtimes` is never deleted .

tools/fuzzer/lldb_eval_libfuzzer_test.cc
```diff
@@ -0,0 +1,163 @@
...
+  std::ostringstream os;
+  os << maybe_expr.value();
+
+  lldb::SBError error;
+  lldb_eval::EvaluateExpression(g_frame, os.str().c_str(), error);
+  return 0;
+}
+
+extern "C" int LLVMFuzzerInitialize(int* argc, char*** argv) {
+  assert(*argc >= 1);
```
werat, line 146: This will fail in `-c opt` node, because the `assert` will be removed and `argc` is no longer used.

tonkosi: Assumed there will be at least one element.

tools/fuzzer/lldb_eval_libfuzzer_test.cc
tonkosi: Done, also turned `g_process` to local variable.

werat (approved):

PR #128, opened 2021-05-20
Prohibit void* subscripting.

PatriosTheGreat:


werat (commented):
Looks great, thanks!

werat:
Not sure if the benchmark fails because you're first-time-contributor or because it's misconfigured... We'll see in the next pull request :)

PR #129, opened 2021-05-21
[fuzzer] Avoid array-to-pointer cast for L-value constraint

tonkosi:
`&(condition ? array : pointer)` isn't valid since `array` is implicitly casted to pointer type, making it an r-value expression.

This change also fixes a subtle assertion failure in `weighted_pick`.


tools/fuzzer/expr_gen.cc
```diff
@@ -1381,7 +1386,7 @@ Enum weighted_pick(
   RealType running_sum = 0;
   for (size_t i = 0; i < array.size(); i++) {
     running_sum += mask[i] ? array[i] : 0;
-    if (choice < running_sum) {
+    if (choice <= running_sum) {
```
tonkosi, line 1389: Note: there's a slight possibility that RNG generates `choice` equal to the total sum, in which none of the options will be selected, resulting in assertion error.

werat:
> Benchmark comparison for [`4a4f97e1`](https://github.com/google/lldb-eval/commit/4a4f97e154dc115e48b7214ad4fec72328e22b19) (base) vs [`3411d271`](https://github.com/google/lldb-eval/commit/3411d27123fc92ed3ca3dd534ebea8f7bff38a06) (PR)
> 
> ```
> BM/ParseInteger                    +0.1257         +0.1258        198730        223720        198696        223698
> ```

Wow these benchmarks are really flaky :(

werat (approved):

PR #132, opened 2021-05-25
[fuzzer] Do not generate void expressions in sizeof

tonkosi:


tools/fuzzer/constraints.h
```diff
@@ -107,6 +107,12 @@ class TypeConstraints {
     return retval;
   }
 
+  static TypeConstraints all_non_void() {
+    TypeConstraints retval = all();
```
werat, line 111: Are there cases when `all()` actually makes sense? All expressions should return something, so we never want `void`, right?

werat (approved):

tools/fuzzer/constraints.h
tonkosi: That's right. Completely replaced `all()` with `all_non_void()`.

PR #137, opened 2021-05-31
Get pointer size from the pointer type object

werat:


tonkosi (approved):
Looks good. I have a question regarding `SBType`s used in parser vs evaluation on value.

lldb-eval/parser.cc
```diff
@@ -2045,8 +2045,7 @@ ExprResult Parser::BuildCStyleCast(Type type, ExprResult rhs,
       }
       // Casting pointer to bool is valid. Otherwise check if the result type
       // is at least as big as the pointer size.
-      // TODO(werat): Use target pointer size, not host.
-      if (!type.IsBool() && type.GetByteSize() < sizeof(void*)) {
+      if (!type.IsBool() && type.GetByteSize() < ctx_->GetPointerSize()) {
```
tonkosi, line 2048: On the evalution side there is an assert where `val.type().GetByteSize()` is used: https://github.com/google/lldb-eval/blob/master/lldb-eval/value.cc#L401. Would it be enough to compare `type.GetByteSize() < rhs_type.GetByteSize()` here, or is `rhs_type` not "tied" to the `target_`?

werat: Actually you're right, we can just compare with `rhs_type.GetByteSize()`, since `rhs_type` is known to be a pointer. At first I thought maybe `std::nullptr_t` can be of a different size, but doc says "`sizeof(std::nullptr_t)` is equal to `sizeof(void *)`".

PR #138, opened 2021-06-02
Implement basic undefined behaviour detection

tonkosi:
This change implements detection of following kinds of undefined behaviour:
 - division/remainder by zero,
 - casting out-of-bounds floating point values to integral types,
 - null pointer arithmetic.

The main purpose is to integrate it in fuzzer, in order to avoid false reports caused by undefined behaviour.

The list of supported UBs is far from complete, but covers all cases encountered so far.


lldb-eval/value.cc
```diff
@@ -420,14 +484,20 @@ Value CastIntegerOrEnumToEnumType(lldb::SBTarget target, Value val, Type type) {
...
+                          Error& error) {
   assert(type.IsEnum() && "target type must be an enum");
   assert(val.type().IsFloat() && "argument must be a float");
 
   llvm::APSInt integer(type.GetByteSize() * CHAR_BIT, !type.IsSigned());
   bool is_exact;
   val.GetFloat().convertToInteger(integer, llvm::APFloat::rmTowardZero,
                                   &is_exact);
+
+  if (IsCastOutOfBounds(val, type)) {
```
werat, line 497: `APFloat::convertToInteger` returns `opStatus` (and has `is_exact` out parameter) -- https://github.com/llvm/llvm-project/blob/c162f086ba632ffaedfe92d63bf21571bc8ae4da/llvm/include/llvm/ADT/APFloat.h#L1113

Could we use it to detect invalid casts?

tonkosi: The `is_exact` wasn't a good choice for floats with non-zero values after decimal point. I haven't checked the `opStatus` before, but it turns out that checking for the `opInvalidOp` flag seems to be enough in this case :)

lldb-eval/ub_detection_test.cc
```diff
@@ -0,0 +1,260 @@
...
+  }
+
+  UbStatus GetUbStatus(const std::string& expr) {
+    auto ctx = lldb_eval::Context::Create(expr, frame_);
+
+    lldb_eval::Error err;
+    lldb_eval::Parser p(ctx);
+    lldb_eval::ExprResult tree = p.Run(err);
+
+    assert(!err && "Error while parsing expression!");
```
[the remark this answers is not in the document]
tonkosi: Problem with `ASSERT_*` macros is that they return `void`, so compiler complains about incompatible type `UbStatus`. `EXPECT_*` macros work, but don't stop on a failed check.

What do you suggest?

lldb-eval/eval.cc
```diff
@@ -917,15 +924,24 @@ Value Interpreter::EvaluateBinaryBitwiseAssign(BinaryOpKind kind, Value lhs,
...
 
 Value Interpreter::PointerAdd(Value lhs, int64_t offset) {
+  if (lhs.GetUInt64() == 0 && offset != 0) {
+    // This is UB and the compiler would generate a warning:
+    //
+    //  warning: performing pointer arithmetic on a null pointer has undefined
+    //  behaviour if the offset is nonzero [-Wnull-pointer-arithmetic]
+    //
+    error_.SetUbStatus(UbStatus::kNullptrArithmetic);
+  }
```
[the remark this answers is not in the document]
tonkosi: I only encountered this with literal zero / `std::nullptr_t` variables with negative offsets. I think this should still be handled in evaluation, and perhaps we could make this more strict (e.g. check for negative offsets instead of non-zero). If we want to limit this only for literal zero / `nullptr`, `is_literal_zero()` method is accessible in `eval.cc` in `AstNode`, but in this case this we first have to iterate through `CStyleCastNode`s before getting to the literal zero node.

This causes issues, here are some examples:
```
> (char*)0 + short_min
lldb value     : `0x00000000ffff8000`
lldb-eval value: `0xffffffffffff8000`

> (char*)null_ptr + short_min
lldb value     : `0x00000000ffff8000`
lldb-eval value: `0xffffffffffff8000`

// No differences:
> (char*)void_null_ptr + short_min
`0xffffffffffff8000`

> (char*)(1-1) + short_min
`0xffffffffffff8000`
```

This kind of UB happens 5-10 times out of ~100000 expressions (which is way less than other kinds of UBs, but still happens sometimes).

tonkosi (commented):
I resolved the float-to-int cast issue, and commented on others. Also decided to use a separate binary for `ub_detection_test`.

lldb-eval/ast.h
```diff
@@ -46,6 +46,12 @@ class AstNode {
   virtual bool is_literal_zero() const { return false; }
   virtual lldb::SBType result_type() const = 0;
 
+  // Is the node literal zero or `std::nullptr_t`, including their casted
+  // versions (e.g. `(int*)nullptr`)?
+  virtual bool is_literal_null() const {
+    return Type(result_type_deref()).IsNullPtrType();
```
werat, line 52: This will return `true` for non-literal as well. Do we want that? I think `return false` should be enough, since all literals are handles it `LiteralNode`.

lldb-eval/eval.cc
```diff
@@ -724,15 +728,25 @@ Value Interpreter::EvaluateBinaryAddition(Value lhs, Value rhs) {
...
+    is_literal_null = is_rhs_null;
   }
   assert(ptr.IsPointer() && "invalid ast: ptr must be a pointer");
   assert(offset.IsInteger() && "invalid ast: offset must be an integer");
+
+  if (is_literal_null && offset.GetInteger().isNegative()) {
+    // Binary addition with null pointer causes mismatches between LLDB and
+    // lldb-eval in some cases.
+    error_.SetUbStatus(UbStatus::kNullptrArithmetic);
+  }
```
werat, line 748: I think this can (and should) be checked "statically" during the parsing. Is there a reason you implemented it in the interperter?

tonkosi: I can't check if the offset is negative in the parser.

lldb-eval/ast.h
tonkosi: Actually the problem isn't with just literal `nullptr`, but any variable of type `std::nullptr_t` (and that can be many different node kinds, e.g. dereference: `*null_ptr_addr`). I could override the method for each of this node kinds separately, but I think this is a simple way to cover all the cases.

Hmm word "literal" is a bit misleading in the method name :/

Also, we only need this for cast nodes (we will never apply binary addition on nullptr or literal zero directly), so maybe we could have `is_casted_null` (or similar name) implemented only in `CStyleCastNode`? E.g. like this:
```cpp
bool CStyleCastNode::is_casted_null() const override {
  Type rhs_type = rhs_->result_type_deref();
  return rhs_type.IsNullPtrType() || rhs_->is_literal_zero() || rhs_->is_casted_null();
}
```

werat (approved):
Awesome, thanks!

PR #140, opened 2021-06-14
Use size_t from the target

tonkosi:


lldb-eval/context.cc
```diff
@@ -81,6 +81,20 @@ lldb::SBType Context::GetBasicType(lldb::BasicType basic_type) {
   return ret;
 }
 
+lldb::SBType Context::GetSizeType() {
+  if (size_type_.IsValid()) {
+    return size_type_;
+  }
+
+  // Get the `size_t` type from the target and cache it for future calls.
+  size_type_ = ctx_.GetTarget().FindFirstType("size_t").GetCanonicalType();
```
werat, line 90: Probably better to use `ResolveTypeByName("::size_t")` here (please check that `::` works as expected in this case). `size_t` can be re-defined in different places and `FindFirstType` will return some random one.

tonkosi: `ResolveTypeByName("::size_t")` seems to work fine on Windows and Linux.

werat (approved):

PR #142, opened 2021-06-18
[fuzzer] Fix Linux specific issues

tonkosi:
Fix issues causing false mismatches on Linux (assuming LLDB 12):
- Use `-static` to build test binaries (this prevents picking up wrong `__log2` from `libm.so`),
- Add option to disable `long double`,
- Take into consideration that lldb-eval doesn't support address-of elision in member-of expressions.


testdata/rules.bzl
```diff
@@ -38,7 +38,7 @@ def binary_gen(name, srcs):
                 platform_opts = "--for-linker -debug:dwarf",
             ),
             "//conditions:default": build_cmd.format(
-                platform_opts = "-lstdc++",
+                platform_opts = "-lstdc++ -static",
```
werat, line 41: Add a comment here about why `-static`

werat (approved):

PR #143, opened 2021-06-23
Use ptrdiff_t from target, fix size_t

tonkosi:
Both `ptrdiff_t` and `size_t` are determined based on OS and architecture. This fixes the previous issue where `size_t` was determined from debug info, which could cause issues if `size_t` was redefined.

Tested on Windows and LInux.


lldb-eval/context.cc
```diff
@@ -86,15 +88,52 @@ lldb::SBType Context::GetSizeType() {
...
+  llvm::Triple triple(llvm::Twine(ctx_.GetTarget().GetTriple()));
+  if (triple.isX86()) {
+    if (triple.isOSWindows() && triple.isArch64Bit()) {
+      ptrdiff_type_ = GetBasicType(lldb::eBasicTypeLongLong);
+    } else {
+      ptrdiff_type_ = triple.isArch64Bit() ? GetBasicType(lldb::eBasicTypeLong)
+                                           : GetBasicType(lldb::eBasicTypeInt);
+    }
+    return ptrdiff_type_;
+  }
```
werat, line 131: I had a glance at definitions for other archs and OS and the Linux-way seems to be the most common. I think we can just do this:
```
if isWindows() {
  return is64bit() ? longlong : int;
} 
return is64bit() ? long : int;
```
(similar with size_t)

werat (approved):

PR #144, opened 2021-06-23
[fuzzer] Integrate new byte reader/writer

tonkosi:


tools/fuzzer/libfuzzer_utils.h
```diff
@@ -135,17 +154,17 @@ class LibfuzzerWriter {
   }
 
   void write_byte(uint8_t byte) {
-    if (size_ < max_size_) {
-      data_[size_++] = byte;
+    if (callback_) {
```
werat, line 157: Is it a valid scenario when `callback_` is not set? If not, let's assert on it's being set.

tonkosi: It should always be set, added assert.

werat (approved):

PR #145, opened 2021-06-25
Fix right shift and detect UB in shift operations

tonkosi:
Use arithmetic right shift for signed values, logical right shift for unsigned values.


werat (approved):
Nice, thanks!

PR #149, opened 2021-07-01
Handle implicit cast to pointers

tonkosi:


lldb-eval/parser.cc
```diff
@@ -2022,6 +2032,14 @@ bool Parser::ImplicitConversionIsAllowed(Type src, Type dst) {
     }
   }
 
+  if (dst.IsPointerType()) {
+    // Literal zero, `nullptr_t` and arrays can be implicitly converted to
+    // pointers.
+    if (is_src_literal_zero || src.IsNullPtrType() || src.IsArrayType()) {
```
werat, line 2038: For the case when src is an array, should we check that the array element type is the same as the pointee type?

tonkosi: Oops, good point, we definitely should!

werat (approved):

PR #151, opened 2021-07-08
Implement `__findnonnull` intrinsic #130

### LINKED ISSUE #130: Implement debugger intrinsic `__findnonnull`
`__findnonnull` is an important intrinsic used by NatVis in the `CustomListItems` element -- https://docs.microsoft.com/en-us/visualstudio/debugger/create-custom-views-of-native-objects?view=vs-2019#customlistitems-expansion

I couldn't find any documentation for it, but empirically it looks like this:

```c++
template <typename T>
int __findnonnull(T** ptr, int64_t size) {
  if (size < 0 || size > 100000000) {
    error("Passing in a buffer size that is negative or in excess of 100 million to __findnonnull() is not allowed.");
  }
  for (int i = 0; i < size; ++ i) {
    if (ptr[i] != nullptr) {
      return i;
    }
  }
  return -1;
}
```

werat:
`__findnonnull` is often used in NatVis for visualizing complex data structures like hash tables. Implement it in a similar way as the native Visual Studio debugger.

Fixes #130


lldb-eval/parser.cc
```diff
@@ -561,6 +531,19 @@ static std::unique_ptr<BuiltinFunctionDef> GetBuiltinFunctionDef(
     return std::make_unique<BuiltinFunctionDef>(identifier, return_type,
                                                 std::move(arguments));
   }
+  if (identifier == "__findnonnull") {
+    lldb::SBType return_type = ctx->GetBasicType(lldb::eBasicTypeInt);
+    std::vector<lldb::SBType> arguments = {
+        // The first argument should actually be "T**", but we don't support
```
tonkosi, line 537: In `eval.cc` it says `T*`. If I understand correctly, it is supposed to accept any pointer type, but the single increment will always be the size of a pointer, instead of `sizeof(T)`?

Seems a bit weird to me, could we add a comment somewhere explaining this?

werat: That's a typo, fixed the comment and added descriptions for the functions.

tonkosi (approved):

PR #152, opened 2021-07-12
Support field access on anonymous structs/unions

### LINKED ISSUE #139: Anonymous structs/unions don't work
Accessing members of anonymous structs/unions doesn't work:

```c+
struct S {
  struct {
    int f;
  };
} s;
s.f = 1;
```

```
> bazel run tools/exec -- "s.f"
...
<expr>:1:3: no member named 'f' in 'S'
s.f
  ^
```

When looking up members we should descent into anonymous members:

> Every member of an anonymous struct is considered to be a member of the enclosing struct or union. This applies recursively if the enclosing struct or union is also anonymous.

tonkosi:
Fixes #139.


werat (approved):
Awesome, looks great!

PR #155, opened 2021-07-14
[fuzzer] Suppress unwanted LSAN reports and improve logging

tonkosi:
- Suppresses LSAN reports causing issues on OSS-Fuzz
- Adds prefix to log messages for easier filtering
- Logs UB statistics


tools/fuzzer/lldb_vs_lldb_eval_libfuzzer_test.cc
```diff
@@ -32,6 +34,40 @@
...
+    return total_ > 0 && total_ % kUbStatsPeriod == 0;
+  }
+
+  size_t count(lldb_eval::UbStatus status) { return count_[status]; }
+
+ private:
+  size_t total_;
+  std::unordered_map<lldb_eval::UbStatus, size_t> count_;
+};
+
```
werat, line 68: This is good approach if things were more complicated, but for now I would opt in for simplicity :0

```
void report_undefined_behaviour(lldb_eval::UbStatus status) {
  static std::unordered_map<lldb_eval::UbStatus, size_t> stats;
  static size_t count;

  stats[status]++;
  count++;

  if (count % kUbStatsPeriod == 0) {
    // inline log_ub_stats here
  }
}
```

tools/fuzzer/lldb_vs_lldb_eval_libfuzzer_test.cc
```diff
@@ -32,6 +34,40 @@
...
+static const char* kLoggingPrefix = "[lldb-eval-fuzzer] ";
+
+// How often to log statistics on UB (e.g. every 500 iterations)?
+static const int kUbStatsPeriod = 500;
+
+extern "C" const char* __lsan_default_options() {
+  return "print_suppressions=0";
+}
+
+extern "C" const char* __lsan_default_suppressions() {
```
werat, line 46: Maybe leave comment or a TODO with a short explanation why we need this and when we can get rid of it

werat (approved):

PR #156, opened 2021-07-20
Handle promotion of bitfields

tonkosi:


werat (approved):
Awesome, thanks for implementing this!

PR #157, opened 2021-07-21
[fuzzer] Tool for filtering libfuzzer logs

tonkosi:


tools/fuzzer/report.py
```diff
@@ -0,0 +1,129 @@
+# Copyright 2021 Google LLC
```
werat, line 1: Add shebang so it can be executed as `./report.py ...`
```suggestion
#!/usr/bin/env python3
# Copyright 2021 Google LLC
```

tools/fuzzer/report.py
```diff
@@ -0,0 +1,129 @@
...
+        disable_ub_stats = True
+      elif arg == "--ub-stats":
+        ub_stats = True
+      # Continue, the argument isn't a file.
+      continue
+
+    for filename in glob.glob(arg):
+      files.append(filename)
+
+  assert (not disable_ub_stats or not ub_stats), "Wrong configuration!"
```
werat, line 120: The script seems quite simple, but even here `argparse` wouldn't be more complicated than manual parsing :)

tonkosi: That's much better, I totally forgot it exists :)

werat (approved):

PR #160, opened 2021-08-13
Fix lookup for static members in value context

tonkosi:


lldb-eval/context.cc
```diff
@@ -271,27 +297,16 @@ lldb::SBValue Context::LookupIdentifier(const std::string& name) const {
...
-
-      if (val_name == name_ref ||
-          val_name == llvm::formatv("::{0}", name_ref).str() ||
-          val_name.endswith(llvm::formatv(" {0}", name_ref).str()) ||
-          val_name.endswith(llvm::formatv("*{0}", name_ref).str()) ||
-          val_name.endswith(llvm::formatv("&{0}", name_ref).str())) {
-        value = val;
-        break;
-      }
+    if (scope_ && !name_ref.contains("::")) {
```
werat, line 300: Should it be possible to lookup static variables from nested types?

```c++
struct Outer {
  struct Inner {
    static constexpr int x = 1;
  };
} foo;

// eval(scope="foo", expr="Inner::x + 1") 
```

tonkosi: Good point! Replaced it with `!global_scope` and added more tests.

werat (approved):
Nice! Thanks for fixing this

PR #161, opened 2021-08-13
Support comparing smart pointers as regular pointers

werat:
LLDB has builtin support for smart pointers from libc++ standard
library. Use it to implement smart-ptr-to-pointer decay in supported
contexts (currently only comparison).


tonkosi (approved):
Looks great!

PR #167, opened 2021-08-20
Implement `reinterpret_cast`

tonkosi:
Explicitly handles `reinterpret_cast` instead of using just C-style cast.

Result of C-style cast to reference types is also fixed to be lvalue instead of rvalue.


werat (commented):
Looks good!

werat (approved):

PR #171, opened 2021-09-17
Improve parsing of simple type specifiers

tonkosi:


werat (commented):
Looks great, thanks for implementing it!

One important addition -- now that `TypeDeclaration` has precise information about the type, we can also improve the resolving part as well. Right now it looks up the type by it's name, however, for basic type there's no need for that. We already know everything we need, so we can just do `Context::GetBasicType`. With this in mind, I think we don't actually need to to-string-conversion logic for basic types at all.

werat:
This PR is improved further in #177

PR #175, opened 2021-09-30
Fix some typos in README

HenryRLee:


README.md
```diff
@@ -5,7 +5,7 @@
 ## What
 
 `lldb-eval` is an [LLDB](https://lldb.llvm.org/)-based library for evaluating
-debug expressions in the context of the debuggee process. All modern debuggers
+debug expressions in the context of the debugger process. All modern debuggers
```
werat, line 8: "debuggee" is correct here, the expressions are evaluated in the context of the "process that is being debugged". Let's just use "debugged", if debuggee looks confusing :)

```suggestion
debug expressions in the context of the debugged process. All modern debuggers
```

werat (changes requested):
Thank you for your contribution!

werat:
disregard the failed benchmark action, it's not working for new contributors unfortunately

HenryRLee:
Thanks for the feedback @werat. I updated the PR accordingly.

werat:
Thanks again for the contribution!

PR #177, opened 2021-11-05
Simplify parsing of simple types specifiers

werat:
based on #171


tonkosi (approved):
Thanks for the update. LGTM with a tiny comment.

PR #183, opened 2021-12-06
Fix evaluation in the reference value context

tonkosi:
Fixes the following when evaluating in a reference value context (e.g. `Object &`):
- `this` keyword,
- static member access (lldb-eval used to look for `Object &::static_member` instead of `Object::static_member`)


lldb-eval/context.cc
```diff
@@ -238,6 +238,10 @@ std::unique_ptr<ParserContext::IdentifierInfo> Context::LookupIdentifier(
     } else {
       // In a "value" scope `this` refers to the scope object itself.
       if (name_ref == "this") {
+        // If `scope_` is a reference, dereference it first.
+        if (scope_.GetType().IsReferenceType()) {
+          scope_ = scope_.Dereference();
+        }
```
werat, line 244: This will modify `scope_` for future calls. If that's okay, should we just do this in the constructor?
Alternatively, don't modify the scope.

tonkosi: Oops, good point. Decided to dereference it in constructor.

werat (approved):

PR #185, opened 2022-01-28
Make interpreter independent of parsing context

tonkosi:
The `Interpreter` used to be dependent on `Context`, which was mostly used for parsing and only very few methods were used by the interpreter.

In this change:
- `Context` is removed from the `Interpreter` (only `clang::SourceManager` is required by the interpreter),
- Parsing and resolving identifiers is improved to allow evaluating the same AST using different evaluation contexts.

In an upcoming PR, new API methods which separate parsing and evaluation are planned to be introduced.


lldb-eval/parser_context.h
```diff
@@ -61,7 +61,7 @@ class ParserContext {
   virtual clang::SourceManager& GetSourceManager() const = 0;
   virtual TypeSP ResolveTypeByName(const std::string& name) const = 0;
   virtual std::unique_ptr<IdentifierInfo> LookupIdentifier(
-      const std::string& name) const = 0;
+      const std::string& name) = 0;
```
werat, line 64: Why did it lose `const`?

lldb-eval/parser_context.h
```diff
@@ -72,6 +72,9 @@ class ParserContext {
   void SetAllowSideEffects(bool allow_side_effects);
   bool AllowSideEffects() const;
 
+  std::tuple<Type::MemberInfo, std::vector<uint32_t>> GetMemberInfo(
+      TypeSP type, const std::string& name);
```
werat, line 76: Can this be `const`? It looks it only needs `GetEmptyType`, which can be make `const` as well.

lldb-eval/eval.cc
```diff
@@ -1184,13 +1244,13 @@ Value Interpreter::EvaluateBinarySubAssign(Value lhs, Value rhs) {
 
   if (lhs.IsPointer()) {
     assert(rhs.IsInteger() && "invalid ast: rhs must be an integer");
-    ret = EvaluateBinarySubtraction(lhs, rhs);
+    ret = EvaluateBinarySubtraction(lhs, rhs, rhs.type());
```
werat, line 1247: Shouldn't the result type here be a pointer? Is there a test for pointer-integer subtract-assign operation?

lldb-eval/eval.cc
```diff
@@ -1184,13 +1244,13 @@ Value Interpreter::EvaluateBinarySubAssign(Value lhs, Value rhs) {
...
   } else {
     assert(lhs.IsScalar() && "invalid ast: lhs must be an arithmetic type");
     assert(rhs.type()->IsBasicType() &&
            "invalid ast: rhs must be a basic type");
     ret = CastScalarToBasicType(target_, lhs, rhs.type(), error_);
-    ret = EvaluateBinarySubtraction(ret, rhs);
+    ret = EvaluateBinarySubtraction(ret, rhs, rhs.type());
```
werat, line 1253: Here `ret.type() == rhs.type()`, right (because we just did the cast)? In other places when choosing between two arguments that are the same, we usually pick the first one (kinda like a "convention").

lldb-eval/eval.cc
```diff
@@ -592,27 +669,7 @@ void Interpreter::Visit(const MemberOfNode* node) {
   // LHS can be a pointer to value, but GetChildAtIndex works for pointers too,
   // so we don't need to dereference it explicitely. This also avoid having an
   // "ephemeral" parent Value, representing the dereferenced LHS.
```
werat, line 671: This comment should be moved into `EvaluateMemberOf`.

lldb-eval/eval.cc
```diff
@@ -269,8 +311,42 @@ void Interpreter::Visit(const LiteralNode* node) {
...
+  switch (identifier.kind()) {
+    using Kind = Context::IdentifierInfo::Kind;
+    case Kind::kValue:
+      val = identifier.value();
+      assert(val.IsValid() && "invalid ast: invalid identifier value");
+      break;
+    case Kind::kContextVar:
+      assert(node->is_context_var() && "invalid ast: context var expected");
+      val = ResolveContextVar(node->name());
+      if (!val.IsValid()) {
```
werat, line 326: This can only happen if the expresion was evaluated with a different set of context variables, right?

werat (commented):
Looks great, only a few minor comments. Thanks!

lldb-eval/eval.cc
tonkosi: Oops, you're right, it should be `lhs.type()`.
We have tests for `pointer -= integer`, but the third argument in `EvaluateBinarySubtraction` is only used for `ptrdiff_t`, so in this case it will be ignored.

lldb-eval/eval.cc
tonkosi: Replaced with `ret.type()`.

lldb-eval/eval.cc
tonkosi: Yes.

Once we split parsing and evaluation, it may also happen that parsing and evaluation parts are provided with the same-name context variables, but different types. Added a TODO to check that.

lldb-eval/parser_context.h
tonkosi: It uses `GetEmptyType()` indirectly, which isn't `const`. I wasn't sure if that was intended or not.
Marked all methods `const`.

lldb-eval/parser_context.h
tonkosi: Done

lldb-eval/eval.cc
tonkosi: Done

werat (approved):

werat:
Awesome, thanks!

PR #186, opened 2022-02-15
Split parsing and evaluation in API

tonkosi:
New API methods are introduced which allow parsing and evaluating expression separately. The main idea is to reuse parsed expressions for multiple evaluations in order to improve performance where appropriate.

Example:
```cpp
struct Data {
  int x;
} a{1}, b{2}, c{3};


lldb::SBType data_type;             // LLDB representation for Data
lldb::SBValue val_a, val_b, val_c;  // LLDB representations for a, b, c

// Parse expression in "type context":
auto compiled_expr = lldb_eval::CompileExpression(target, data_type, "x * x", ...);

// Evaluate in different "value contexts":
auto a_squared = lldb_eval::EvaluateExpression(val_a, compiled_expr);  // gives 1
auto b_squared = lldb_eval::EvaluateExpression(val_b, compiled_expr);  // gives 4
auto c_squared = lldb_eval::EvaluateExpression(val_c, compiled_expr);  // gives 9
```


testdata/test_binary.cc
```diff
@@ -217,6 +217,7 @@ static void TestMemberOfInheritance() {
   Parent* parent = &obj;
 
   // BREAK(TestMemberOfInheritance)
+  // BREAK(TestSomething)
```
werat, line 220: Leftover?

lldb-eval/context.cc
```diff
@@ -224,7 +229,12 @@ std::unique_ptr<ParserContext::IdentifierInfo> Context::LookupIdentifier(
   // Lookup context variables first.
   auto context_var = context_vars_.find(name);
   if (context_var != context_vars_.end()) {
-    return IdentifierInfo::FromContextVar(context_var->second);
+    return IdentifierInfo::FromValue(context_var->second.inner_value());
+  }
+
+  auto context_arg = context_args_.find(name);
+  if (context_arg != context_args_.end()) {
+    return IdentifierInfo::FromContextVar(context_arg->second);
```
werat, line 237: Maybe rename it to `FromContextArg` for consistency with other namings?
Or maybe even better to `FromType` for symmetry with `FromValue`.

lldb-eval/context.h
```diff
@@ -119,7 +119,8 @@ class Context : public ParserContext {
   }
   lldb::SBExecutionContext GetExecutionContext() const { return ctx_; }
 
-  void SetContextVars(std::unordered_map<std::string, TypeSP> context_vars);
+  void SetContextVars(std::unordered_map<std::string, Value> context_vars);
```
werat, line 122: The evaluator doesn't use `Context` now, so there should never be `context_vars` if the parser only operates with types, right?
We've discussed the optimization of "using the values directly if they're available", but I think for the first approach it's better to drop it for simplicity.

lldb-eval/api.h
```diff
@@ -73,6 +98,43 @@ LLDB_EVAL_API
...
+LLDB_EVAL_API
+ParsedExpression ParseExpression(lldb::SBFrame, const char* expression,
+                                 lldb::SBError& error);
+
+LLDB_EVAL_API
+ParsedExpression ParseExpression(lldb::SBFrame, const char* expression,
+                                 Options opts, lldb::SBError& error);
+
+LLDB_EVAL_API
+ParsedExpression ParseExpression(lldb::SBTarget target, lldb::SBType scope,
```
werat, line 110: Can we return a `std::shared_ptr<ParsedExpression>` instead? Then the caller can just pass it around and not worry about the lifetime.

lldb-eval/api.h
```diff
@@ -73,6 +98,43 @@ LLDB_EVAL_API
...
+                                 const char* expression, lldb::SBError& error);
+
+LLDB_EVAL_API
+ParsedExpression ParseExpression(lldb::SBTarget target, lldb::SBType scope,
```
werat, line 114: Let's call it `CompileExpression`, we're getting closer and closer to mini-compiler here :)
The return value can be `CompiledExpr` for shortness.

lldb-eval/api.h
```diff
@@ -52,11 +55,33 @@ struct ContextVariableList {
...
 struct Options {
   bool allow_side_effects = false;
+  ContextArgumentList context_args = {};
   ContextVariableList context_vars = {};
 };
 
+struct ParsedExpression {
+  std::shared_ptr<SourceManager> source;
+  std::unique_ptr<AstNode> tree;
+  lldb::SBTarget target;
```
werat, line 81: Do we need the target here? It can always be extracted from the value, right?

werat (commented):
Overall looks great! I have only a couple of suggestions about the naming and simplifying context variable handling.

lldb-eval/api.h
tonkosi: Yes, it can be extracted from the value/frame. Removed.

lldb-eval/api.h
tonkosi: Done, makes more sense.

lldb-eval/api.h
```diff
@@ -73,6 +98,43 @@ LLDB_EVAL_API
...
+ParsedExpression ParseExpression(lldb::SBTarget target, lldb::SBType scope,
+                                 const char* expression, Options opts,
+                                 lldb::SBError& error);
+
+LLDB_EVAL_API
+lldb::SBValue EvaluateParsedExpression(ParsedExpression& expression,
+                                       lldb::SBError& error);
+
+LLDB_EVAL_API
+lldb::SBValue EvaluateParsedExpression(ParsedExpression& expression,
```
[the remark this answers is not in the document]
tonkosi: Replaced with `std::shared_ptr<CompiledExpr>`

lldb-eval/context.cc
tonkosi: Replaced with `FromContextArg` (to avoid confusion since there already is `FromThisKeyword` which takes a single type as parameter).

lldb-eval/context.h
tonkosi: Changed `Context` to operate with context arguments only.

testdata/test_binary.cc
tonkosi: Oops, yes.

werat (approved):
awesome! 👍

PR #191, opened 2022-03-08
Add character parsing capabilities

shyamliagrawal:
Used [CharLiteralParser](https://clang.llvm.org/doxygen/classclang_1_1CharLiteralParser.html)  to parse characters.


lldb-eval/eval_test.cc
```diff
@@ -3405,3 +3405,30 @@ TEST_F(EvalTest, TestRegistersNoDollar) {
...
+
+  EXPECT_THAT(Eval("1 + U'A'"), IsEqual("66"));
+  EXPECT_THAT(Eval("U'B' - U'A'"), IsEqual("1"));
+  EXPECT_THAT(Eval("U'A' == U'B'"), IsEqual("false"));
+  EXPECT_THAT(Eval("U'A' == U'A'"), IsEqual("true"));
+
+  EXPECT_THAT(Eval("1 + L'A'"), IsEqual("66"));
+  EXPECT_THAT(Eval("L'B' - L'A'"), IsEqual("1"));
+  EXPECT_THAT(Eval("L'A' == L'B'"), IsEqual("false"));
+  EXPECT_THAT(Eval("L'A' == L'A'"), IsEqual("true"));
```
werat, line 3428: Maybe also try some non-ascii values? E.g. non-latin characters or emojis.

lldb-eval/parser.cc
```diff
@@ -137,6 +137,21 @@ std::tuple<lldb::BasicType, bool> PickIntegerType(
   return {lldb::eBasicTypeUnsignedLongLong, true};
 }
 
+lldb::BasicType PickCharType(clang::CharLiteralParser& literal) {
```
werat, line 140: I think all the methods below are `const`, so this can be passed via a const-ref. Generally we avoid non-const references as much as possible, as it's very easy to accidentally mutate the parameter.

```suggestion
lldb::BasicType PickCharType(const clang::CharLiteralParser& literal) {
```

lldb-eval/parser.cc
```diff
@@ -2133,6 +2154,36 @@ ExprResult Parser::ParseBooleanLiteral() {
...
+  llvm::SmallString<8> CharBuffer;
+  bool CharInvalid = false;
+  llvm::StringRef ThisTok = pp_->getSpelling(token_, CharBuffer, &CharInvalid);
+  clang::CharLiteralParser charLiteral(ThisTok.begin(), ThisTok.end(), loc,
+                                       *pp_, token_.getKind());
+
+  if (CharInvalid || charLiteral.hadError()) {
+    BailOut(
+        ErrorCode::kInvalidCharLiteral,
+        "Failed to parse token as char-constant: " + TokenDescription(token_),
```
werat, line 2171: Use `llvm::formatv()` to format an error message.

lldb-eval/parser.cc
```diff
@@ -2133,6 +2154,36 @@ ExprResult Parser::ParseBooleanLiteral() {
...
 }
 
+ExprResult Parser::ParseCharLiteral() {
+  ExpectOneOf(clang::tok::char_constant, clang::tok::wide_char_constant,
+              clang::tok::utf16_char_constant, clang::tok::utf32_char_constant);
+  clang::SourceLocation loc = token_.getLocation();
+
+  llvm::SmallString<8> CharBuffer;
+  bool CharInvalid = false;
+  llvm::StringRef ThisTok = pp_->getSpelling(token_, CharBuffer, &CharInvalid);
```
werat, line 2164: There is an overload that just takes token and returns `std::string`. Is there a reason to use this one instead?

lldb-eval/parser.cc
```diff
@@ -2133,6 +2154,36 @@ ExprResult Parser::ParseBooleanLiteral() {
...
+  clang::CharLiteralParser charLiteral(ThisTok.begin(), ThisTok.end(), loc,
```
werat, line 2165: We follow Google C++ style guide and use `camel_case` for local variables -- https://google.github.io/styleguide/cppguide.html#Variable_Names

lldb-eval/parser.cc
```diff
@@ -2133,6 +2154,36 @@ ExprResult Parser::ParseBooleanLiteral() {
...
+
+  auto basicType = PickCharType(charLiteral);
+  llvm::APInt literal_value(
+      ctx_->GetBasicType(basicType)->GetByteSize() * CHAR_BIT,
+      charLiteral.getValue());
+
+  ConsumeToken();
+  return std::make_unique<LiteralNode>(
+      loc, ctx_->GetBasicType(basicType), literal_value,
+      /*is_literal_zero*/ literal_value.isNullValue());
```
werat, line 2184: Does 0-char count as a literal zero? Assigning or comparing against `'\0'` doesn't work:
```
int* x = 0;
bool b = x == '\0';

error: comparison between pointer and integer ('int *' and 'char')
    bool b = x == '\0';
             ~ ^  ~~~~
```

lldb-eval/parser.cc
```diff
@@ -2133,6 +2154,36 @@ ExprResult Parser::ParseBooleanLiteral() {
...
+    BailOut(
+        ErrorCode::kInvalidCharLiteral,
+        "Failed to parse token as char-constant: " + TokenDescription(token_),
+        token_.getLocation());
+    return std::make_unique<ErrorNode>(ctx_->GetEmptyType());
+  }
+
+  auto basicType = PickCharType(charLiteral);
+  llvm::APInt literal_value(
+      ctx_->GetBasicType(basicType)->GetByteSize() * CHAR_BIT,
```
werat, line 2178: `ctx_->GetBasicType(basicType)` is relatively cheap, but still it's better to resolve it only once -- save it into a variable.

lldb-eval/parser_context.h
```diff
@@ -11,6 +11,7 @@ enum class ErrorCode : unsigned char {
   kOk = 0,
   kInvalidExpressionSyntax,
   kInvalidNumericLiteral,
+  kInvalidCharLiteral,
```
werat, line 14: Unfortunately adding new error codes is a breaking API change and it will break the Visual Studio extension :(
Let's use `kInvalidNumericLiteral` for now and leave a TODO, please, saying that it should be a separate error code.

lldb-eval/parser.cc
```diff
@@ -137,6 +137,21 @@ std::tuple<lldb::BasicType, bool> PickIntegerType(
...
+  } else if (literal.isAscii()) {
+    return lldb::eBasicTypeChar;
+  } else if (literal.isWide()) {
+    return lldb::eBasicTypeWChar;
+  } else if (literal.isUTF16()) {
+    return lldb::eBasicTypeChar16;
+  } else if (literal.isUTF32()) {
+    return lldb::eBasicTypeChar32;
+  }
+  return lldb::eBasicTypeChar;
```
werat, line 152: Is there a valid use case when we get here? It looks like all valid options are covered above, so we can return `lldb::eBasicTypeInvalid` instead.

testdata/test_binary.cc
```diff
@@ -1236,6 +1240,7 @@ void main() {
   RegisterCtx rc;
   rc.TestRegisters();
 
+  TestCharParsing();
```
werat, line 1243:
```suggestion
  TestCharParsing();

```

werat (commented):
Thanks, looks good! I have a few comments about the style and some questions, but generally the approach is correct, nice work.

lldb-eval/eval_test.cc
tonkosi: Maybe also try invalid literals (e.g. `''`) and multi-char literals with more than 4 chars?

shyamliagrawal: Added tests for non-ascii and multi-char literals. For '', the error we get in lldb-eval is different from lldb.

lldb-eval/eval_test.cc
```diff
@@ -3405,3 +3405,43 @@ TEST_F(EvalTest, TestRegistersNoDollar) {
...
+
+  EXPECT_THAT(Eval("1 + 'ABC'"), IsEqual("4276804"));
+  EXPECT_THAT(Eval("'ABD' - 'ABC'"), IsEqual("1"));
+  EXPECT_THAT(Eval("'ABC' == 'BCD'"), IsEqual("false"));
+  EXPECT_THAT(Eval("'ABC' == 'ABC'"), IsEqual("true"));
+
+  EXPECT_THAT(Eval("U'ъ'== U'猫'"), IsEqual("false"));
+  EXPECT_THAT(Eval("U'ü'== U'ü'"), IsEqual("true"));
+  EXPECT_THAT(Eval("U'u'== U'ü'"), IsEqual("false"));
+  EXPECT_THAT(Eval("'ABCDE'=='ABCDF'"), IsEqual("false"));
```
werat, line 3446: Why is this `false`? Is it undefined behavior?

lldb-eval/eval_test.cc
```diff
@@ -3405,3 +3405,43 @@ TEST_F(EvalTest, TestRegistersNoDollar) {
...
+  EXPECT_THAT(Eval("'A' == 'B'"), IsEqual("false"));
+  EXPECT_THAT(Eval("'A' == 'A'"), IsEqual("true"));
+
+  /*Following test fails detecting different results from
+   *as u8 prefix is not handled in lldb
+   *EXPECT_THAT(Eval("1 + u8'A'"), IsEqual("66"));
+   *EXPECT_THAT(Eval("u8'B' - u8'A'"), IsEqual("1"));
+   *EXPECT_THAT(Eval("u8'A' == u8'B'"), IsEqual("false"));
+   *EXPECT_THAT(Eval("u8'A' == u8'A'"), IsEqual("true"));
+   */
```
werat, line 3421: Use single line comments -- `//`, multiline are generally more difficult to edit.

lldb-eval/eval_test.cc
werat, line 3446:
```suggestion
  EXPECT_THAT(Eval("U'ъ' == U'猫'"), IsEqual("false"));
  EXPECT_THAT(Eval("U'ü' == U'ü'"), IsEqual("true"));
  EXPECT_THAT(Eval("U'u' == U'ü'"), IsEqual("false"));
  EXPECT_THAT(Eval("'ABCDE' == 'ABCDF'"), IsEqual("false"));
```

lldb-eval/eval_test.cc
werat: > For '', the error we get in lldb-eval is different from lldb.

Different how? Can we match the error message?

werat (commented):
Good to go modulo a couple of nits, thanks!

lldb-eval/eval_test.cc
shyamliagrawal: In lldb-eval, it fails with error: Unexpected token. 
For lldb, it shows error: expected expression 
                and warning: empty char constant

werat (approved):
Thanks, looks great!

PR #192, opened 2022-05-09
Add special support for string-literals

tonkosi:
Works only if the expression consists of a single string-literal token. Allocates memory for the string value locally (i.e. not for the process being debugged).
For other cases parsing results in "not implemented" error.


werat (approved):
Looks great, thanks!

PR #199, opened 2022-11-22
Fixes required for targeting wasm32

pfaffe:
Drive-by: Fix pointer comparison on 32bit targets

### DIFF
--- a/lldb-eval/eval.cc
+++ b/lldb-eval/eval.cc
@@ -984,8 +984,9 @@ Value Interpreter::EvaluateComparison(BinaryOpKind kind, Value lhs, Value rhs) {
   // Must be pointer/integer and/or nullptr comparison.
   size_t ptr_size = target_.GetAddressByteSize() * 8;
 
-  bool ret = Compare(kind, llvm::APSInt(lhs.GetInteger().trunc(ptr_size), true),
-                     llvm::APSInt(rhs.GetInteger().trunc(ptr_size), true));
+  bool ret =
+      Compare(kind, llvm::APSInt(lhs.GetInteger().sextOrTrunc(ptr_size), true),
+              llvm::APSInt(rhs.GetInteger().sextOrTrunc(ptr_size), true));
   return CreateValueFromBool(target_, ret);
 }
 
--- a/lldb-eval/eval.cc
+++ b/lldb-eval/eval.cc
@@ -982,7 +982,7 @@ Value Interpreter::EvaluateComparison(BinaryOpKind kind, Value lhs, Value rhs) {
   }
 
   // Must be pointer/integer and/or nullptr comparison.
-  size_t ptr_size = target_.GetAddressByteSize();
+  size_t ptr_size = target_.GetAddressByteSize() * 8;
 
   bool ret = Compare(kind, llvm::APSInt(lhs.GetInteger().trunc(ptr_size), true),
                      llvm::APSInt(rhs.GetInteger().trunc(ptr_size), true));
--- a/lldb-eval/eval.cc
+++ b/lldb-eval/eval.cc
@@ -981,6 +981,11 @@ Value Interpreter::EvaluateComparison(BinaryOpKind kind, Value lhs, Value rhs) {
   }
 
   // Must be pointer/integer and/or nullptr comparison.
+  if (target_.GetAddressByteSize() == 4) {
+    bool ret = Compare(kind, static_cast<uint32_t>(lhs.GetUInt64()),
+                       static_cast<uint32_t>(rhs.GetUInt64()));
+    return CreateValueFromBool(target_, ret);
+  }
   bool ret = Compare(kind, lhs.GetUInt64(), rhs.GetUInt64());
   return CreateValueFromBool(target_, ret);
 }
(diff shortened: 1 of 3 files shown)


lldb-eval/parser.cc
```diff
@@ -142,7 +142,7 @@ std::tuple<lldb::BasicType, bool> PickIntegerType(
 lldb::BasicType PickCharType(const clang::CharLiteralParser& literal) {
   if (literal.isMultiChar()) {
     return lldb::eBasicTypeInt;
-  } else if (literal.isAscii()) {
+  } else if (literal.isOrdinary()) {
```
werat, line 145: I fixed this in main branch while fighting presubmits, please rebase.

lldb-eval/eval.cc
```diff
@@ -981,6 +981,11 @@ Value Interpreter::EvaluateComparison(BinaryOpKind kind, Value lhs, Value rhs) {
   }
 
   // Must be pointer/integer and/or nullptr comparison.
+  if (target_.GetAddressByteSize() == 4) {
+    bool ret = Compare(kind, static_cast<uint32_t>(lhs.GetUInt64()),
+                       static_cast<uint32_t>(rhs.GetUInt64()));
+    return CreateValueFromBool(target_, ret);
+  }
```
werat, line 988: Hm, does `GetUInt64()` return incorrect values for 4-byte pointers (i.e. with garbage in the other 4 bytes)? I think this is not the only place where we use `GetUInt64()` to get the "pointer value" from the variable, it should probably be handled in other places as well.

lldb-eval/eval.cc
```diff
@@ -981,7 +981,10 @@ Value Interpreter::EvaluateComparison(BinaryOpKind kind, Value lhs, Value rhs) {
   }
 
   // Must be pointer/integer and/or nullptr comparison.
-  bool ret = Compare(kind, lhs.GetUInt64(), rhs.GetUInt64());
+  size_t ptr_size = target_.GetAddressByteSize();
+
+  bool ret = Compare(kind, lhs.GetInteger().trunc(ptr_size),
+                     rhs.GetInteger().trunc(ptr_size));
```
werat, line 987: I think `APInt::trunc(width)` takes the width in _bits_ -- https://github.com/llvm/llvm-project/blob/6cac3dc9db57cbf5f9f9092d97f878be3db5020d/llvm/include/llvm/ADT/APInt.h#L1242
This probably works by accident :P

lldb-eval/eval_test.cc
```diff
@@ -2093,43 +2268,52 @@ TEST_F(EvalTest, TestReferenceScope) {
...
   EXPECT_THAT(Eval("bf.a"), IsEqual("1023"));
   EXPECT_THAT(Eval("bf.b"), IsEqual("9"));
   EXPECT_THAT(Eval("bf.c"), IsEqual("false"));
   EXPECT_THAT(Eval("bf.d"), IsEqual("true"));
+#ifndef __EMSCRIPTEN__
   EXPECT_THAT(Scope("bf").Eval("a"), IsEqual("1023"));
   EXPECT_THAT(Scope("bf").Eval("b"), IsEqual("9"));
   EXPECT_THAT(Scope("bf").Eval("c"), IsEqual("false"));
   EXPECT_THAT(Scope("bf").Eval("d"), IsEqual("true"));
+#endif
```
werat, line 2283: What's wrong with these? Is scoped evaluation not supported for `wasm32`?

lldb-eval/eval_test.cc
```diff
@@ -2698,14 +2907,18 @@ TEST_F(EvalTest, TestPostfixIncDec) {
...
+  EXPECT_THAT(EvalWithContext("$p--", vars_),
+              IsEqual(Is32Bit() ? "0x0000000c" : "0x000000000000000c"));
   EXPECT_THAT(EvalWithContext("$p-- - 1", vars_),
-              IsEqual("0x0000000000000004"));
+              IsEqual(Is32Bit() ? "0x00000004" : "0x0000000000000004"));
+#endif
 }
 
+#ifndef __EMSCRIPTEN__
 TEST_F(EvalTest, TestDereferencedType) {
```
werat, line 2922: What's wrong with this test? It seems to me this should work for `wasm32` too...

lldb-eval/eval_test.cc
```diff
@@ -2819,8 +3038,10 @@ TEST_F(EvalTest, TestCompositeAssignmentInvalid) {
   ASSERT_TRUE(CreateContextVariable("$i", "1"));
   EXPECT_THAT(EvalWithContext("($i += 1) -= 2", vars_),
               IsError("side effects are not supported in this context"));
+#endif
 }
 
+#ifndef __EMSCRIPTEN__
 TEST_F(EvalTest, TestCompositeAssignmentAdd) {
```
werat, line 3045: Have you considered doing something like this when disabling the whole test?

```
#ifdef __EMSCRIPTEN__
  // Doesn't work on wasm32 because of X
  GTEST_SKIP() << "not supported on wasm32";
#else
```

lldb-eval/eval_test.cc
pfaffe: It's the unconditional comparison with lldb's output here. wasm32 can't run lldb, so we can't do this comparison.

lldb-eval/eval_test.cc
pfaffe: I did, but ifdef'ing out the entire test simplifies the wasm wrapper, so that it doesn't have to provide noop implementations of things like EvalWithContext() or Scope(). This seemed cleaner than implementing methods that aren't supposed to be called.

werat (approved):

